PR #22, opened 2015-02-21
Added onlineplayers

xCausxn:
Added onlineplayers which utilizes the length of the players array to
make it always available.

Possible to add bots to this count but i think just players would be
best.

Updated readme to include this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Otherwise, the returned object is guaranteed to contain the following keys:
 * **name**
 * **map**
 * **password**: Boolean
+* **onlineplayers**
 * **maxplayers**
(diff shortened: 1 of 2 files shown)


sonicsnes:
I don't think I'm going to accept this. The readme clearly states that players.length is the suggested way to obtain this information, and adding another field is both information duplication (a source of errors and confusion in the future), as well as ambiguous (ie. does it or does it not include bots?)

xCausxn:
I made it only count the players as it is was it done on gametracker. However adding the bots would be easy anyway :P 

I thought it would be one less thing people would have to do :)

PR #25, opened 2015-03-29
Make quake3 have a numplayers variable too

youssefdetovernickr:
When you query a quake 3 server (like Wolfenstein Enemy Terroritory) it basically gets the player names, but it does not have a raw.numplayers variable. With this it's added :)

### DIFF
--- a/protocols/quake3.js
+++ b/protocols/quake3.js
@@ -6,14 +6,14 @@ module.exports = require('./quake2').extend({
 	},
 	finalizeState: function(state) {
 		state.name = this.stripColors(state.name);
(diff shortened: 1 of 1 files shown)


sonicsnes:
state.players.length is the proper way to query for this information. This is mentioned briefly in the README: "It can usually be assumed that the number of players online is equal to the length of the players array. Some servers may return an additional player count number, which may be present in the unstable raw object."

The only reason that some servers have a player count in the raw object, is that they provide this number separately in the data they return. This raw value should not be used in most cases.

PR #35, opened 2015-10-26
Adjusted Port for Rust to remove offset by 1.

lukeacl:
The query port was changed to the game port instead of being offset by
1 as mentioned at http://playrust.com/devblog-74/. Duplicated the line
and renamed the old to rustlegacy for backwards support.


squarefanatic:
Hi there! Just following up if you'll be pulling this one in.

PR #70, opened 2017-12-14
Dynamic port binding

Coriou:
I wanted to run GameDig in parallel but was unable to because the socket created always binds to the same hardcoded port.

Now it will use a new and free UDP port for each query.

I'm using [udp-free-port](https://github.com/mcollina/udp-free-port) to find a free port.


Coriou:
Fuck sorry I pushed a commit with that I didn't want to push here but only use for myself..

sonicsnes:
Seems like there is a lot going on in this code review that might be unnecessary.

(1) Pulling the global timeout out of the options and passing them into the protocol constructors, then putting them back into the options is not needed and is a lot of code duplication.
(2) gamedig can already support parallel queries made within a single daemon -- there's no reason to make a separate socket for every request.
(3) It's true that gamedig currently doesn't work from multiple independent daemons at the same time due to the hard-coded port. I honestly don't recall why it was originally written this way (that was a long while back!), but there shouldn't be any need to use udp-free-port to find a fresh port. Since the udp port is being fully bound and not just used for sending, it should be sufficient to just [remove the 21943 parameter](https://github.com/sonicsnes/node-gamedig/blob/4c18b6a3fdb2c6f137664985f27efa8fbc38ec43/lib/index.js#L9), since node will [automatically bind a random port for you](https://nodejs.org/api/dgram.html#dgram_socket_bind_port_address_callback).

Coriou:
Thanks for your reply, it's a messy PR I'll start again in a much cleaner fashion soon.

However, a couple of remarks / questions :

(1) For the timeout option feature, it was only meant for me I didn't want to include it in this pull request. I do believe that it'd however be good to have, I'll see if I'll try to implement it in a clean way and open a new PR. If you have any suggestions on how to implement it at "API level" (ie. the query() method) I'd love to take your advice

(2) I haven't found out how parallel queries can be achieved with the current implementation (without having multiple sockets) - if you can explain how this can be achieved maybe this PR is useless all together

(3) While it's true that Node will rely on the OS to try to bind to a new random port to a socket, it does not guarantee that the port will be free and will throw an error if it isn't. The free-udp-port package essentially starts and binds a new socket randomly and if it errors assumes the binding didn't work and then tries again until binding is successful. It's a little overhead indeed, but it seems - to me at least, with my current understanding - that it's the safest way to always have a socket binded on a free random port

I'll close this PR and I'll wait for your reply to start re-implementing this small change in a clean branch (if necessary). Same for the timeout option really, which I'd really need and think should be implemented in GameDig's core.

Coriou:
Me again.

Not only was my code sloppy, but my whole understanding of the project was kinda wrong, borderline stupid.

I opened two new PRs (#71 & #72), each implementing the two features discussed above in a much simpler and straightforward way.

Coriou:
Hey @sonicsnes, any chance you'll merge the PRs about this ? Thanks

PR #72, opened 2017-12-18
Timeout option

Coriou:
Can now set a "timeout" option at query level (in milliseconds). Will be applied to both TCP & UDP requests.

Is useful for servers located far away where the default 1 second timeout is too low (in my case, Australian server can take up to 1.5 seconds to reply).

Example :

`GameDig.query({ type: 'XXX', host: 'xxx.xxx.xxx.xxx', port: 123, timeout: 1000})`

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ function(e,state) {
 * **type**: One of the game IDs listed in the game list below
 * **host**
 * **port**: (optional) Uses the protocol default if not set
+* **timeout**: (optional) The max timeout, in milliseconds, for both UDP & TCP requests (default: 1000)
 * **notes**: (optional) Passed through to output
 
 ### Return Value
--- a/lib/index.js
+++ b/lib/index.js
@@ -54,6 +54,12 @@ class Gamedig {
             query.udpSocket = udpSocket;
             query.type = options.type;
 
+            if ( options.timeout ){
+                query.options.tcpTimeout = options.timeout
+                query.options.udpTimeout = options.timeout
+                delete options.timeout
+            }
+
             if(!('port' in query.options) && ('port_query' in query.options)) {
                 if(Gamedig.isCommandLine) {
                     process.stderr.write(
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "gamedig",
-	"version": "0.2.30",
+	"version": "1.0.37",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {


omarAjmi:
what about the CLI command version? same attributes?

Coriou:
Haven't tried the CLI version, but based on the code at `bin/gamedig.js` it should work the same yes. If you try, let me know if it worked or not, I'll look into it and fix it if doesn't

sonicsnes:
Sorry about the delay, suffice to say... I've been a bit busy 😄 

Anyways, this option should actually already exist, I believe you can already pass udpTimeout and tcpTimeout as parameters to your query.

sonicsnes:
Closing this as it appears tcpTimeout and udpTimeout already do this.

Coriou:
@sonicsnes `tcpTimeout` and `udpTimeout` do that indeed, but most people will not know if they're using the UDP or TCP protocol, why make it so complicated and undocumented ? It's beyond me.

The hardcoded timeout is 1 second only, not 10, and in my testing it's too short for a lot of servers.

You've closed all the pull requests, this is the only one you cared to comment so - at least - thanks for that.

Any reason why you refuse to provide more options to the users of this library ? Any reason people should continue using this branch as you don't seem to care to maintain it anymore ? 

It took you over a month to look at them, and you just closed them without a word. I don't get it

sonicsnes:
@Coriou All of the other merge requests were accepted with tweaks. They are all in the latest published revision of gamedig. I think that calling the option `timeout` is a bit misleading, because setting it to 1000 will not guarantee that the request returns within one second. For instance, many of the protocols have five or more udp round trips, and could thus take up to five seconds to execute with such a setting.

Another option would be to remove the udp and tcp timeouts entirely, forcing everything to get caught by the global 10 second timeout, but this would result in extremely slow queries for offline servers.

sonicsnes:
At a minimum, I agree that it may be reasonable to combine tcpTimeout and udpTimeout into a single option (socketTimeout?) and mention the ability to increase it in the README, but its default value ultimately has to be a balance between slow queries for legimiately offline servers and failed queries for slow servers.

Coriou:
Agreed for `socketTimeout`, it's more obvious indeed.

> but its default value ultimately has to be a balance between slow queries for legimiately offline servers and failed queries for slow servers.

Yeah you're right. I wouldn't even change the default of 1 second for the socket to timeout. I don't have actual stats or anything on how this library is used, but I would bet that most people's use case don't require timeouts to be longer than 1 second. Just that the option is needed in some cases (like mine : I live in Europe and querying Australian servers often take more than a second).

Coriou:
Great

PR #76, opened 2018-01-25
Log the time a request takes to complete

Coriou:
Extends the "query" object of the response to include "Time taken" which is populated by the time (in seconds) it took to complete the request - won't be present when query fails


sonicsnes:
This is now returned as `duration` in the unstable query object, along with the other query-process related outputs.

PR #91, opened 2018-10-19
On error return options within the callback

mtchtech:
Gives back more information when promise errors out.

Currently using this in a series of server information queues, without giving the options back or more information it's hard to tell what server isn't responding. I figure this is a simple addition to the callback that otherwise wouldn't pass in any information about the query before.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -89,7 +89,7 @@ class Gamedig {
             if(callback.length === 2) {
                 promise
                     .then((state) => callback(null,state))
-                    .catch((error) => callback(error));
+                    .catch((error) => callback(error, options));
             } else if (callback.length === 1) {
                 promise
                     .then((state) => callback(state))


sonicsnes:
The promise shouldn't need to return the options back to you if you've passed them in. You already control the javascript context, so you should be able to keep your own options variable available to yourself in the callback.

sonicsnes:
The promise shouldn't need to return the options back to you if you've passed them in. You already control the javascript context, so you should be able to keep your own options variable available to yourself in the callback.

mtchtech:
hmm, maybe this is one of those javascript things that I don't fully understand. I can add a few snippets of what i'm doing with the query. Essentially I am querying 18 servers over 36 seconds. Some of which only give the information "timeout" as a string, meaning after a few retries (probably spanning around a minute or more). The only way I could figure out which game server (tf2) was actually timing out was passing in the options to the callback. 

used code within a node.js bot:
```javascript
server = {
    name:"Test Server 1",
    address:"xxxx:27024"
};
Gamedig.query({
    type:'tf2',
    host:serverProp[0],
    port:serverProp[1],
    notes:server,
    socketTimeout: 10000,
    maxAttempts: 3
},
function(e,state) {
    if(e) {
        var serverId = state.notes.name;
        console.log(state.notes.name + " : " + JSON.stringify(e));
        //removed irrelevant code
    } else {
        //removed more code
    }
});
```

PR #96, opened 2018-11-14
Fixed crash on invalid port range.

xcpep:
Fixes a crash that occurs when the port is out of the specified range. I think the source of this problem could be the port offset, but may also be servers reporting an invalid port to the Steam masterlist. Regardless, a nice fix to a uncommon issue.

### DIFF
--- a/protocols/core.js
+++ b/protocols/core.js
@@ -286,7 +286,11 @@ class Core extends EventEmitter {
             console.log(this.options.address+':'+this.options.port_query+" UDP-->");
             console.log(HexUtil.debugDump(buffer));
         }
-        this.udpSocket.send(buffer,0,buffer.length,this.options.port_query,this.options.address);
+        if(this.options.port_query > 0 && this.options.port_query <= 65535) {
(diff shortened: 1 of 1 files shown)


bbashy:
Like this crash?

```
dgram.js:436
    throw new ERR_SOCKET_BAD_PORT(port);
    ^

RangeError [ERR_SOCKET_BAD_PORT]: Port should be >= 0 and < 65536. Received 65536.
    at Socket.send (dgram.js:436:11)
    at Valve._udpSendNow (/usr/local/lib/node_modules/gamedig/protocols/core.js:289:24)
    at process.nextTick (/usr/local/lib/node_modules/gamedig/protocols/core.js:267:18)
    at internalTickCallback (internal/process/next_tick.js:70:11)
    at process._tickCallback (internal/process/next_tick.js:47:5)
    at Function.Module.runMain (internal/modules/cjs/loader.js:778:11)
    at startup (internal/bootstrap/node.js:300:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:826:3)
```

xCausxn:
Although this fix will work, I personally feel it would be better suited in the Gamedig class as a validation step.

PR #102, opened 2019-01-17
Support generating Markdown from JSON

pizzafox:
There doesn't seem to be any beneficial reason to using a text file over JSON. It's difficult for GameDig to use internally because it's not machine readable. You made your own parser for it to convert it into JSON, ultimatly.

Switching to JSON would also be simple to turn into Markdown, you could do it in a few lines instead of several files. It would also help developers like me who want access to a machine readable list of every game's properties.

I've written the code for parsing the JSON and generating Markdown, the next step is to rewrite the internal functions in `GameResolver`. There's only one comment in it so it's very difficult to understand what's happening for me. I'll need someone who's more experienced with that to help.


sonicsnes:
Hi!

While I truly do appreciate the contribution, I'm not entirely sure I see the benefit here. I do agree that JSON is more standard, but the parser for games.txt is only 40 lines and results in a much more concise (and in my opinion, human-readable at a glance) file format for this specific information. JSON increases the line count of that file by nearly 10x, and doesn't support inline comments.

If you require pre-parsed access to this information, you can always get it from `(new GameResolver()).games` (although I certainly can't guarantee this api will remain forever!)

If you have a good argument for why this information should be more accessible to clients, I'd like to hear it and perhaps we could solve the access solution in a more appropriate way.

Thanks!,
~ SonicSNES

PR #131, opened 2019-07-01
Update games.txt

WinkelB:
Added Support for Game Squad


sonicsnes:
Looking at https://squad.gamepedia.com/Server_Configuration#Ports_to_Open , it seems like the default query port for this game is actually 27165. Testing with a couple servers on public lists, this seems to be the case.

TheSpartaPT:
@sonicsnes merge this and change it to the default port, it's annoying that I have to add squad manually

PR #132, opened 2019-07-12
Fix: (node:73) Warning: a promise was rejected with a non-error: [object String]

TemaSM:
- [x] Fixed warning about rejecting with non-error (was spamming in console)
`(node:73) Warning: a promise was rejected with a non-error: [object String]`
- [x] Fixed `package.json` indentation (tabs -> spaces), so now it looks better even in GitHub source viewer
- [x] Regenerated `package-lock.json` just for better freshness

I'm going to send another PRs soon (maybe few days or even weeks), because I think we should remove bloated in size `request` package from dependencies, standardize indentation style and code itself.
But for now, just waiting for your review or merging @sonicsnes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,662 +1,635 @@
 {
-	"name": "gamedig",
-	"version": "2.0.13",
-	"lockfileVersion": 1,
-	"requires": true,
-	"dependencies": {
-		"@types/cheerio": {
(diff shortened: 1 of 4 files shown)


sonicsnes:
Cherry-picked the incorrect error throwing part of this. Thanks for the contribution!

Issue #156, opened 2020-01-03
Add discord widget API


Douile:
Would like to add a discord query to go along with teamspeak and mumble.

This is possible to do without a discord API key by using the widget API, however it requires servers to enable the widget option in settings.

Since I couldn't find documentation on this API here is a [link to a public example](https://discordapp.com/api/guilds/309408516857266178/widget.json)

It would involve a simple HTTP get request followed by JSON parsing.

API URL: `https://discordapp.com/api/guilds/{GUILDID}/widget.json`

Only issue is that servers cannot be addressed with IPs only discord snowflake IDs however I don't think this would be too much of a problem.

_I may make a pull request to add this_

mmorrisontx:
Merged. Thank you!

PR #157, opened 2020-01-04
Add discord protocol

Douile:
Add the discord protocol as described in #156 
Tested locally ✔️ 
I had to override DNSResolver as the address parsed is a discord ID and not a hostname.
I also changed the bin file as when I was passing snowflake IDs minimist was parsing them as integers, which would later cause errors (I just forced _ argument to be string).


mmorrisontx:
Whoops, sorry I missed this PR! ... for a year.

Cleaned up a bit and released in 3.0.4.

BTW, I noticed your `unessential-requests` branch, this shouldn't be needed anymore since we made similar changes to the valve protocol in the 3.0 release :)

PR #163, opened 2020-02-15
Fix wrong player count in minecraftvanilla protocol

### LINKED ISSUE #162: Getting minecraft playercount always returns double of the actual playercount
Today I noticed that for minecraft using `players.length` always returned the wrong playercount.
I found out that it is always the actual playercount multiplied by 2 and found the source of the problem:
https://github.com/sonicsnes/node-gamedig/blob/master/protocols/minecraftvanilla.js#L50-L61
I guess you forgot to wrap the for loop in an else? @sonicsnes

P4sca1:
Closes #162

### DIFF
--- a/protocols/minecraftvanilla.js
+++ b/protocols/minecraftvanilla.js
@@ -57,7 +57,7 @@ class MinecraftVanilla extends Core {
             }
         }
 
-        // players.sample may not contain all players or no players at all, depending on how many players are only.
+        // players.sample may not contain all players or no players at all, depending on how many players are online.
         // Insert a dummy player object for every online player that is not listed in players.sample.
         // Limit player amount to 10.000 players for performance reasons.
         for (let i = state.players.length; i < Math.min(json.players.online, 10000); i++) {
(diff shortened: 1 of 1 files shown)


protocols/minecraftvanilla.js
```diff
@@ -54,9 +54,10 @@ class MinecraftVanilla extends Core {
                     name: player.name
                 });
             }
-        }
-        for (let i = 0; i < Math.min(json.players.online, 10000); i++) {
-            state.players.push({});
+        } else {
+            for (let i = 0; i < Math.min(json.players.online, 10000); i++) {
```
sonicsnes: Ah, I believe the original intention here was for i to be initialized to state.players.length. `sample` often doesn't include a full player list, so this loop is meant to fill out the rest of the slots with placeholders.

P4sca1: Ah okay, I thought that when sample exists it always includes all players and otherwise we need to use placeholders.

P4sca1: @sonicsnes I pushed another commit and also added coments to make this more clear.

P4sca1:
@sonicsnes Anything else needed before releasing this fix?

PR #175, opened 2020-04-24
Add support for Post Scriptum

arkuar:


xCausxn:
May be worth using the `squad` type as PostScriptum is running on the same framework. Which fixes the password check.

arkuar:
Good point.

Douile:
Should also add beyond the wire, I believe it also uses squad.

PR #187, opened 2020-08-01
Change singleton structure to reduce singleton instantiation complexity.

inf3rnus:
Cleaned up the singleton class so that unneeded methods are removed. This should not be a breaking change because the change originates from the index.js file.

However, if you have something to test this against that'd be great.

Cheers ! 🚀

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,23 +1,14 @@
 const QueryRunner = require('./QueryRunner');
 
-let singleton = null;
-
 class Gamedig {
     constructor() {
         this.queryRunner = new QueryRunner();
     }
-
     async query(userOptions) {
         return await this.queryRunner.run(userOptions);
     }
-
-    static getInstance() {
-        if (!singleton) singleton = new Gamedig();
-        return singleton;
-    }
-    static async query(...args) {
-        return await Gamedig.getInstance().query(...args);
-    }
 }
 
-module.exports = Gamedig;
+const gameDig = new Gamedig();
+Object.freeze(gameDig);
+module.exports = gameDig;


inf3rnus:
Tested on the minecraft server us.mineplex.com, seems to work... We should build out some tests for this bad boy...

sonicsnes:
Originally, the plan was to support separate instances of the Gamedig query object. For instance, if the user wanted to use a different outgoing UDP port for each one. I'm not sure it was ever publicized, but I'm not sure we can say for sure that nobody is using
```
const Gamedig = require('gamedig');
const gamedig = new Gamedig();
gamedig.query(...);
```

mmorrisontx:
Originally, the plan was to support separate instances of the Gamedig query object. For instance, if the user wanted to use a different outgoing UDP port for each one. I'm not sure it was ever publicized, but I'm not sure we can say for sure that nobody is using

```
const Gamedig = require('gamedig');
const gamedig = new Gamedig();
gamedig.query(...);
```

inf3rnus:
Hmmmmm, I do see your point about other developers using the 'new' keyword in their code to explicitly instantiate their own instance vs the implicit instantiation offered by the static query method, although I think that because of the repo docs, that situation is unlikely (but I understand the desire to not break things).

With that said, would you find it helpful if the Gamedig class were changed to have a factory method for creating new QueryRunner objects, with all of the modules it depends on being changed to include the option to specify an outgoing port via the call to udpSocket.bind(); (line 17 of GlobalUdpSocket.js)?

Alternatively, and I think this fits the style of the project better, a new option could be added to the query() methods of the Gamedig class called 'outgoingPort' and the Gamedig class could internally keep a map of every outgoing port and a QueryRunner instance. Then, whenever query() is called, it checks to see if the 'outgoingPort' option is present, and if it is, checks the map and uses that QueryRunner instance if it's present. If it's not present, it proceeds as usual using the default port assigned by the OS. However, if the instance for a specified outgoingPort is not instantiated yet, (QueryRunner would need to be refactored ever so slightly), but the 'outgoingPort' could be passed in as an argument to its constructor, and then passed to GlobalUdpSocket's constructor, where the port # would then be passed to udpSocket.bind()(line 17 of GlobalUdpSocket.js)

Idk, would love to hear your thoughts, I think this project is awesome.

mmorrisontx:
Closing since it's unlikely we'll go in this direction. `listenUdpPort` is now an option on constructed Gamedig instances, so if anything, we'll probably move in that direction rather than moving closer to singletons.

PR #257, opened 2021-12-02
Refactoring, regression fix, lowering the minimum required nodejs version to 8

a-sync:
The main goal of this PR is to bring back support for nodejs 8+.  
Other then the _got_ dependency's version and the new syntax used in Results.js, i found no problem in running gamedig with nodejs 8.4. (older got has the same interface, and works well atm. for gamedig)  

 * refactor Results.js to nodejs 8 compatible syntax
   * fix regression caused by Results and Players
     * line [R22](https://github.com/gamedig/node-gamedig/pull/257/files#diff-b9b84a84c9bc70636f693af0ef1fd9bd20398478e396010f83994870bbc71e0dR22)
     * line [R47](https://github.com/gamedig/node-gamedig/pull/257/files#diff-b9b84a84c9bc70636f693af0ef1fd9bd20398478e396010f83994870bbc71e0dR47)
 * revert dependency _got_ to v9
 * lower the minimum required nodejs version to 8
 * run `npm audit fix`


cetteup:
I can't help but wonder: Why add support for a nodejs version that reached EOL in early 2020?

a-sync:
I can only speak from personal experience:
 1. it is a lot easier to update a dependency of a deployed app then the deployment environment
     (so when a protocol changes and gamedig needs to be updated, i am not presented with extra tasks (sometimes even out of my control))
 2. i don't see any real reason to drop support since the two things preventing it (mentioned above) are not a requirement for any protocol (but correct me if i'm wrong)

mmorrisontx:
While I appreciate the contribution, I have no intention of making this package compatible with a version of node that is past its EOL. ESPECIALLY if it requires downgrading `got` to an older version.

Newer gamedig releases will take advantage of more node 12 quality of life features, not less.

Although you do have me curious about which hosting provider only supports versions of node which are totally obsolete.

PR #281, opened 2022-05-17
Add npx usage in the README

rainst:
I often use gamedig just like that, I thought it would be useful to have it in the readme :)


mmorrisontx:
Thanks!

PR #291, opened 2022-06-30
Update games.txt

Bizow:
Add V Rising


Bizow:
protocol-valve will work for V Rising

PR #293, opened 2022-07-09
chore(deps): remove moment

c43721:
This PR removes `moment` and replaces it with `date-fns`

- [`date-fns` repo](https://date-fns.org/)
- [`moment` repo](https://momentjs.com/docs/#/-project-status/)

Since moment is no longer maintained beyond security patches, I took the opportunity to remove it in favor of `date-fns` which is a moment-like alternative. One could argue to use the [temporal api](https://tc39.es/proposal-temporal/docs/) which is in stage 3 of the ecmascript proposal, so if that is desired I can swap over to Temporal instead.

The benfits of date-fns is that it has tree-shaking by default, so we're not importing a whole library. With that, it's also being actively maintained :). We also have types built in!

Please let me know if I should migrate to Temporal instead.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,597 +1,1355 @@
 {
-	"name": "gamedig",
-	"version": "4.0.3",
-	"lockfileVersion": 1,
-	"requires": true,
-	"dependencies": {
-		"@sindresorhus/is": {
-			"version": "4.6.0",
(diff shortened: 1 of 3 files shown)


mmorrisontx:
I'm amazed this is the only use of moment. We could probably just rip it out and use `date.toISOString().split('T')[0]`

c43721:
That's an excellent plan. Will do.

Issue #294, opened 2022-07-19
CSGO reports "Max Players" as a player object


nickdnk:
Hello

It seems that somehow GameDig thinks there's always a player on the server called "Max Players", meaning you can't count on this array to be empty if the server is empty. Doesn't make a lot of sense to me.

```js
{
  name: 'Counter-Strike: Global Offensive',
  map: 'workshop/377312340/aim_deagle',
  password: true,
  raw: {
    protocol: 17,
    folder: 'csgo',
    game: 'Counter-Strike: Global Offensive',
    appId: 730,
    numplayers: 0,
    numbots: 0,
    listentype: 'd',
    environment: 'l',
    secure: 1,
    version: '1.38.3.7',
    tags: [ 'empty', 'secure' ]
  },
  maxplayers: 10,
  players: Players(1) [ Player { name: 'Max Players', raw: [Object] } ], // ???
  bots: Players(0) [],
  connect: 'docker-host:20015',
  ping: 0
}
```

Output from `users` command in console:
```
users
<slot:userid:"name">
0 users
```

mmorrisontx:
This issue was fixed in gamedig aaaaaages ago here: https://github.com/gamedig/node-gamedig/blob/master/protocols/valve.js#L229

No idea why csgo throws this object into their player array for no reason, but they do.

Are you running a 5 year old release? 😁

mmorrisontx:
Additionally, if you want full player name info, don't forget about the cvar you need to enable server-side: https://github.com/gamedig/node-gamedig#counter-strike-global-offensive

nickdnk:
This is on version 4.0.4, so no, it's not an old release.

Looking at the code, it looks for a property called `steamappid`, while the one the server actually sends is called `appId`, so the code you linked to is never going to work.

mmorrisontx:
Whoooops, yep that would be a problem.

nickdnk:
https://github.com/gamedig/node-gamedig/pull/296

nickdnk:
Cool. Thanks.

nickdnk:
Changelog update missing btw.

PR #295, opened 2022-07-19
Fix wrong filter for "Max Players" for CSGO

nickdnk:
https://github.com/gamedig/node-gamedig/issues/294


nickdnk:
Also added CSGO to `AppId` enum. Don't know if you want this, but it seems like it's being used like that for other games.

PR #303, opened 2022-10-02
use minecraftbedrock protocol for minecraft bedrock

matthi4s:


mmorrisontx:
The gamedig minecraft protocol uses minecraftbedrock as one of its data sources. The reason it's collapsed into one is mostly because users often don't know the difference between minecraft versions... Is it not working properly as written in some use case?

matthi4s:
Thanks for the explanation, the Bedrock resolution didn't work properly for me which might be caused by something else at that moment. I got the response from a Vanilla/Java server back instead which was unexpected when explicitly querying Bedrock. It's also an unnecessary overhead to query all types if the user knows the correct protocol.

Maybe adding an explicit `minecraftbedrock` and `minecraftjava` type and adding a note to the README would solve this?

cetteup:
If you are unaware: You can define the protocol directly by passing `protocol-[prococol name]` as the game.

matthi4s:
Thanks, I didn't know that. That works for me, maybe a note in the README would still be helpful.

podrivo:
We currently have a **Minecraft: Bedrock Edition** protocol. It was working great on the server I've tested!

Docs are also being updated here https://github.com/gamedig/node-gamedig/pull/480.

CosminPerRam:
Closed as we now have a separate Minecraft: Bedrock Edition entry (`mbe`).

Issue #306, opened 2022-12-13
Resolving to IPv6 and receiving send EINVAL error


SmollClover:
Hi, I'm currently using Gamedig for a Project to read out a Status of a FiveM Server but Gamedig has started to resolve the Host of the Gameserver to the IPv6 which then Errors out.
On my Local machine it resolves to the IPv4 of the Server and works fine.

Server output
```
Q#0 Starting
Q#0 Protocol: FiveM
Q#0 Options: {
  socketTimeout: 2000,
  attemptTimeout: 10000,
  maxAttempts: 1,
  port: 30120,
  protocol: 'fivem',
  requestRules: false,
  type: 'fivem',
  host: 'fivem.modern-gaming.net',
  debug: true
}
Q#0 DNS Lookup: fivem.modern-gaming.net
Q#0 Standard Resolve: fivem.modern-gaming.net
Q#0 Found address: 2a01:4f8:c012:c6b8::1
2a01:4f8:c012:c6b8::1:30120 UDP(51193)-->
Buffer length: 16 bytes
ff ff ff ff 67 65 74 69 6e 66 6f 20 78 78 78 00 
            g  e  t  i  n  f  o     x  x  x     

Q#0 Query failed with error Error: send EINVAL 2a01:4f8:c012:c6b8::1:30120
    at doSend (node:dgram:705:16)
    at defaultTriggerAsyncIdScope (node:internal/async_hooks:463:18)
    at afterDns (node:dgram:651:5)
    at process.processTicksAndRejections (node:internal/process/task_queues:83:21)
Error: Failed all 1 attempts
    at QueryRunner.run (/root/.local/share/nvm/v18.1.0/lib/node_modules/gamedig/lib/QueryRunner.js:95:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Gamedig.query (/root/.local/share/nvm/v18.1.0/lib/node_modules/gamedig/lib/index.js:11:16)
Attempt #1 - Port=30120 Retry=0:
Error: send EINVAL 2a01:4f8:c012:c6b8::1:30120
    at doSend (node:dgram:705:16)
    at defaultTriggerAsyncIdScope (node:internal/async_hooks:463:18)
    at afterDns (node:dgram:651:5)
    at process.processTicksAndRejections (node:internal/process/task_queues:83:21)
```

Is there a way to force it to resolve to IPv4?

cetteup:
[...]

PR #307, opened 2022-12-13
Allow direct control of IP family to be returned by DNS lookup

cetteup:
As mentioned in #306, query may currently fail if the DNS lookup happens to return an IPv6 `AAAA` record instead of an existing IPv4 `A` record. In order to avoid such issues, these changes enables users to directly control which type of record/ip family they accept.

Tested locally with the server mentioned in #306.


mmorrisontx:
Released in 4.0.6

Issue #310, opened 2023-01-09
Minecraft always reporting 0 ping regression.


WhyKickAmooCow:
There is a regression for the bug reported #233 where the querying Minecraft always returns a result of zero ping. This is because in Core.runOnce, ping is set using shortestRTT which is always 0 as the resolver does not use registerRtt. This overrides the ping the resolver set.

The solution is (as the Minecraft appears to be the only resolver setting state.ping manually) to use registerRtt in the Minecraft resolver, rather than setting the ping value itself.

Issue #233, opened 2021-06-15
Minecraft - Ping is always 0


Emiliaaah:
When using the type minecraft it always returns 0.
But raw.vanilla.ping (in my case) does return the ping and not 0.

mmorrisontx:
Thanks for the report, definitely a bug.

AlexandreSama:
The bug is still there, but this is really weird ! I give you a screenshot about results

![Capture d’écran 2021-08-11 215136](https://user-images.githubusercontent.com/56075082/129094019-5c0e8a8d-6251-4e4d-a013-dc5ab292e1f2.png)

mmorrisontx:
For now, you can pull it from one of the raw subquery objects, unless someone wants to contribute a MR for pulling it into the merged response :)

WhyKickAmooCow:
It looks like there may have been a regression, this still seems to be affecting me on version 4.0.5. Exactly the same symptoms in the return value, raw.vanilla.ping is still accurate but ping is always 0.

PR #311, opened 2023-01-09
Fixed Minecraft resolver only reporting 0 ping. #310

WhyKickAmooCow:
Minecraft resolver now uses registerRtt rather than setting state.ping. This prevents the ping value being overwritten with 0 as shortestRTT is never set.

This fixes the bug reported in #310, which is a regression of #233.


mmorrisontx:
Released in 4.0.6

Issue #312, opened 2023-01-09
Add Risk of Rain 2 support.


CosminPerRam:
Risk of Rain 2 is a game developed by Hopoo Games.
Steam page [here](https://store.steampowered.com/app/632360/Risk_of_Rain_2/).
Steam ID: 632360

The game uses the Valve Query Protocol, the default game connection port is 27015 with the query port being + 1 than the game connection port, so its 27016.

The game also reports a "Test": "Value" server rule on every server rules query.

PR #316, opened 2023-01-11
Add support for Risk of Rain 2

cetteup:
Last one for today. Requested by @CosminPerRam in #312.

### DIFF
--- a/games.txt
+++ b/games.txt
@@ -217,6 +217,7 @@ rtcw|Return to Castle Wolfenstein|quake3|port_query=27960
 rfactor|rFactor|rfactor|port=34397,port_query_offset=-100
 ricochet|Ricochet|valve|port=27015
 riseofnations|Rise of Nations|gamespy1|port_query=6501
+ror2|Risk of Rain 2 (2020)|valve|port=27015,port_query_offset=1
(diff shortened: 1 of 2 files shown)


CosminPerRam:
Great! I haven't been looking into this repository implementation on how it handles games and stuff (as I have added support for the game in my own rust crate). I've seen it isn't supported here and thought to open an issue about it as it uses the standard valve query (altough, there is that 'test': 'value' rule (shall we remove it?)).

games.txt
```diff
@@ -217,6 +217,7 @@ rtcw|Return to Castle Wolfenstein|quake3|port_query=27960
 rfactor|rFactor|rfactor|port=34397,port_query_offset=-100
 ricochet|Ricochet|valve|port=27015
 riseofnations|Rise of Nations|gamespy1|port_query=6501
+ror2|Risk of Rain 2 (2020)|valve|27015,port_query_offset=1
```
mmorrisontx, line 220: should probably be `port=27015,port_query_offset=1` instead of `27015,port_query_offset=1`

cetteup: Well spotted! Fixed.

PR #327, opened 2023-01-18
Just adding Warning and tips for ValveProtocol

DoctorWhoFR:
I am currently using gamedig for most of my games servers, i just find some new comportment on getting ValveProtocol game info.

By adding some time in the socketTimeout, i prevent all error returned.

This is a little but really working solution in fact ^^


CosminPerRam:
Sometimes you do need higher `socketTimeout`, but this is already documented in the README, here:
https://github.com/gamedig/node-gamedig/blob/ff5f6ef2e29c3185f1aee526f88a104055272421/README.md?plain=1#L44-L45
Adding another reminder only on the Valve Protocol would be redundant, unless this is a repeating problem.

PR #362, opened 2023-09-16
Add support for running using deno

Douile:
Prefix node imports with "node:" and gate a socket API that is not implemented in [deno](https://deno.land) so that the library can be used there. This should not break node and doesn't in my brief testing.


CosminPerRam:
Thanks for contributing to Node-GameDig!
Although I'm not sure how many Deno users are for now, I do think (If not breaking stuff occurs) that this would be fitting considering the small amount of changes for future use.
I will also test some things before merging.

CosminPerRam:
After merging, please make an issue to track [denoland/deno#20138](https://github.com/denoland/deno/issues/20138) to make the needed changes (hopefully just to remove that `if` statement) when stuff gets fixed.

bin/gamedig.js
```diff
@@ -65,4 +67,10 @@ gamedig.query(options)
 
             printOnPretty({error: error});
         }
+    })
+    .then(() => {
+      // https://github.com/denoland/deno/issues/20138
+      if (typeof Deno !== "undefined") {
+        gamedig.queryRunner.udpSocket?.socket?.close();
+      }
```
CosminPerRam, line 75: I dont exactly understand something, `unref` is not implemented, but why do we close the connection only here on the binary one? wouldn't we need to also close instances after being used in lib-only?

Douile: Yes, sorry I should've mentioned that. I just wasn't sure how to do that on the library as there's no real close method. I guess maybe could close after every query and reconstruct it for the next.

CosminPerRam:
Hey, coming back late to this (sorry!), could you rebase please? Also, please add a line to the changelog mentioning support for Deno.

Douile:
:cry: https://github.com/gamedig/node-gamedig/commit/93a9095d9999165c3388ed03de56c4a0f8f20c79 changed line endings to CRLF breaking all my diffs.

CosminPerRam:
Sorry about that, could you re-implement?

PR #407, opened 2023-11-19
Remove games.txt and replace it with an in-code solution

CosminPerRam:
We would define a game in the `games.txt` file in the following format: `id | pretty name for readme | protocol | options | extra`.

Storing this data in a file then reading it turned out to be problematic, there has been many cases where this caused issues, such as using this in NextJS, where you would need this hack around it: 
```js
const config = {
    experimental: {
        serverComponentsExternalPackages: ["gamedig"],
    },
};
```
This was also inefficient, although it would be read only once (and therefore be lazily loaded), this wasn't exactly a good practice and wouldn't provide any realistic advantages instead of just having them stored in a code file.

One advantage (at least, kind of?) would have been readability, as a game was defined in one line, it was easy to see its properties, but this would get janky with longer lines and not exactly the best for contributors.

This PR removes said game definitions file and replaces it with an in-code solution, in theory it will speed up only the first lookup but reduce memory usage (EDIT: actually not really as I think they would be referenced in said array, not deep copied), as before the game could have been found in 2 places, the `gamesByKey` object and the `games` array (which this one was used just to automatically generate the `GAMES_LIST.md` file).


Douile:
It would be nice if the changelog mentioned the removal of `GameResolver`. Also It's removal will require updating the typescript type definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/gamedig/lib/GameResolver.d.ts

CosminPerRam:
Sure thing, removal of `GameResolver` has been added to changelog in `f472d8b0e4d39415daf98c49232c153a5aaf9431`.

PR #410, opened 2023-11-19
dns: Use node's built in isIP check instead of regex

### LINKED ISSUE #408: bug: DNS resolver isIP detection doesn't detect IPv6 addresses (and could succeed for invalid addresses)
**Describe the bug**
In DNSResolver there is a "fast" path that detects whether a given hostname is already an IP,  this uses a regex.

The regex fails to detect IPv6 addresses meaning if the user passes one an unnecessary DNS query will always happen.

The regex used for IPv4 detection doesn't truly match how a IPv4 address cab be represented. While this doesn't cause bad behaviour as an invalid IPv4 address cannot be a valid domain name, I think the current regex could be vulnerable to a regex DOS attack as it doesn't put upper bounds on repetition lengths.

**Steps To Reproduce**
- Provide an IPv6 address and observe DNS query being made.

- Provide an invalid IPv4 address and observe it gets passed directly to connect: `25555.1299999.24444.1`

**Expected behavior**
We should detect IPv6 addresses, and use an efficient algorithm for checking both.

**Additional context**
node:net provides an isIP function that is probably more robust than using our own regex (and also detects IPv6).

Douile:
Fixes #408 by replacing the old IP check with node's built-in check (compatible with Deno).

The IPv6 part of this is mostly irrelevant as I couldn't find any IPv6 game servers and the CLI doesn't handle IPv6 address properly anyway, but I left it in for future proofing.

### DIFF
--- a/lib/DnsResolver.js
+++ b/lib/DnsResolver.js
@@ -1,4 +1,5 @@
 import dns from 'node:dns'
+import { isIP } from 'node:net'
 import punycode from 'punycode/punycode.js'
 
 export default class DnsResolver {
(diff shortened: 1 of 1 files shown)


CosminPerRam:
Thanks!

PR #412, opened 2023-11-20
fixed 7 Days to die ID

jammsen:
Fixed not working sd2d to sdtd, which is working


CosminPerRam:
Oh, seems I forgot the `2`, thanks for spotting this out.
Maybe I should set up an CI that regenerates this file on each `games.js` change, we'll avoid all of these problems.

jammsen:
No problem and yes a CI would be good for that. Agreed.

PR #428, opened 2023-12-03
Support for ElDewrito

Sphyrna-029:
Added support for monitoring Halo Online (ElDewrito) servers. Added proper return object and raw data.


CosminPerRam:
Please make sure the origin is also `4.3.0` (:

PR #435, opened 2023-12-12
tools: Add find ID changes to list changes to game IDs

Douile:
This tool iterates through the commit history finding commits where lib/games.js has been modified. Whenever it has it loads the before/after state and compares them to generate a list of IDs that were removed/added/changed (changed is based on comparing the name of the game).

The output of this script can then be converted into an object mapping between old IDs and new IDs using a jq one-liner.


CosminPerRam (approved):
Thanks, comes in handy!

PR #448, opened 2024-01-02
chore: extract dayz from valve

CosminPerRam:
DayZ runs on Valve A2S but it is modified quite a bit and their data types needs to be read differently than standard ones, this PR extracts everything dayz related into another protocol file (which still runs on top of the valve protocol file), clears up a lot the cluttered Valve protocol file.

### DIFF
--- a/lib/games.js
+++ b/lib/games.js
@@ -654,7 +654,7 @@ export const games = {
     options: {
       port: 2302,
       port_query_offset: 24714,
-      protocol: 'valve'
+      protocol: 'dayz'
     }
   },
   dayzmod: {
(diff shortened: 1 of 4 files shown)


CosminPerRam:
another possible thing to do would be to still detect DayZ in the Valve protocol (via its app id) and instead of returning the valve response, run the dayz protocol, but I dont really know a reason to do this, better yet as said in my previous comment, its now possible to get the valve-only response with no dayz data addons, if you want the addons, run the DayZ protocol.

podrivo:
This is incredible @CosminPerRam! Thank you! 🙌 🚀

podrivo:
I was wondering, will this be part of the version 4.3.2?

CosminPerRam:
That specific commit no, that's changed behaviour regarding how players (or bots) are added into the list and that's a breaking change, so its for 5.X

PR #449, opened 2024-01-02
fix(valve): dont skip players with no name

CosminPerRam:
Commit 90e26bb and PR #389 removed the behaviour of adding players with empty data on the `numplayers` count, the valve protocol skipped adding players if their names are empty, some games (such as DayZ) provide players data (time played) but no names. Its better to have partial data than no data at all (beforehand this PR and the mentioned one, the query would return empty players data).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ see next point) and `tshock` (which is `terraria`).
 
 #### Games
 * Removed the players::setNum method, the library will no longer add empty players as 
-a placeholder in the `players` fields.
+placeholders in the `players` fields.
+* Valve: dont skip players with no name and keep state.raw.players.
 * Stabilized field `numplayers`.
 * BeamMP (2021) - Added support.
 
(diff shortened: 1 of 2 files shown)


podrivo:
Hello @CosminPerRam! Since we're going to keep players with no name, I'm interested in using this raw data, but it's currently being sorted with bots and deleted.

Do you think we can keep the `state.raw.players` and don't delete it? It would be really useful to have the raw data.

https://github.com/gamedig/node-gamedig/blob/fccd61c4eaac9a80a1f07021f04435fbc182832d/protocols/valve.js#L277

CosminPerRam:
For sure, thanks for the idea!

podrivo:
Awesome! Thank you so much! 🙌

podrivo:
@CosminPerRam Just realized that `state.raw.players` is still being deleted here:

https://github.com/gamedig/node-gamedig/blob/0b8ebc871f8986a61ad0ea43190a98ecc5e68d83/protocols/valve.js#L280-L282

Will work on a solution for us and create a new PR.

PR #454, opened 2024-01-11
fix: wrong port attempts order

### LINKED ISSUE #451: bug: maxAttempts firing all attempts, regardless
**Describe the bug**
I'm testing some querying for DayZ servers and sometimes it can take a little longer to get a response, or sometimes the response is not complete, etc, and this is normal, just like DayZ and mods work. So too avoid having to manually retry, I've set the `maxAttempts: 5` and notice that it will fire all 5 attempts, regardless. I've updated to `maxAttempts: 20` and it took forever, and it went through all 20. If I set to `maxAttempts: 1` and the server response is good, it will work on the first try, but adding `maxAttempts` will just increase time.

Feels like it's a bug, where it will repeat all attempts, but only the last one will successfully fire.

**Steps To Reproduce**
I'm testing this via JS in a Express app, using this code:

```js
let state = await gamedig.query({
    type: 'dayz',
    host: server.ip,
    port: server.port_query,
    requestRules: true,
    maxAttempts: 5,
    debug: true
})
```

**Expected behavior**
If I'm adding `maxAttempts: 20`, I was expecting to get a successful response maybe on the 3rd call or maybe the 7th, but it goes through all 20, every time.

I'm not sure this is related to DayZ, because I've only tested with this game. Also, not sure if this a feature or a bug.

Thank you! 🙏

CosminPerRam:
The user given port should always be the first one, try the others next.
~~Closes #451.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Made the library a `module`.
 * Removed `GameResolver`, moved the `GameDig` class in a separate file.
 * Modified exports, now the library exports `games` and `protocols` alongside the `GameDig` class.
+* The user-given port is always attempted as the first once.
 
 #### Games
 * Removed the following alternative game ids: `minecraftping` (as it was deprecated and the same thing as `minecraft`),
--- a/lib/QueryRunner.js
+++ b/lib/QueryRunner.js
@@ -45,13 +45,13 @@ export default class QueryRunner {
     }
 
     if (userOptions.port) {
+      attempts.push(optionsCollection)
+
       if (!userOptions.givenPortOnly) {
         if (gameQueryPortOffset) { addAttemptWithPort(userOptions.port + gameQueryPortOffset) }
 
         if (userOptions.port === gameOptions.port && gameQueryPort) { addAttemptWithPort(gameQueryPort) }
       }
-
-      attempts.push(optionsCollection)
     } else if (gameQueryPort) {
       addAttemptWithPort(gameQueryPort)
     } else if (gameOptions.port) {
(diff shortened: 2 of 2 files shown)


CosminPerRam:
This could pose some issues:
- Users don't usually know the exact query port of a game and this PR regresses in performance for these: one additional port to be tested first
- Also quite a big breaking change, as many code bases could do this, they'll regress upon migrating to 5.0.0

But could improve on performance:
- For those who do or for the games that their connect port is the same as the query port: skips up to 2 possible ports

I think I'll need to do some research to check on how many games provide that additional one or two port attempts and see whats the balance.

cetteup:
Regarding the performance impacts: Users who know and are confident about the query port can already reduce the number of attempted ports to 1 using the `givenPortOnly` option. That should not change. But, in my opinion, it allows performance optimization towards other users who aren't sure about the query port. The assumption being that if you know the query port, you should always provide `givenPortOnly` to avoid additional queries (e.g. in case the server is offline).

Maybe the attempt procedure needs to be documented more clearly, I don't actually know how many people really use/know about `givenPortOnly` and it's effects.

P.S.: I really appreciate the effort you are putting into the library. Making your own code changes via PRs makes them very easy to follow. And the draft PRs mean I can start thinking about changes long before they get merged. 🙂

GuilhermeWerner:
It has occurred to me a few times to realize that the lib was querying another port before trying the one I provided.

> I don't actually know how many people really use/know about `givenPortOnly` and it's effects.

It turns out that I didn't know about this option, I'm going to activate it, because in my use case I always provide the correct port, and I should have performance gains with that.

cetteup:
In my understanding, `givenPortOnly` is not impacted at all. The option makes gamedig really only try the port you provide. Any order of (other) ports to try becomes completely irrelevant. It does not add any ports either before the one you specified (current behavior) or after the one you specified (proposed new behavior). But either way the option ensures that your "ports to try" list will have a length of exactly one.

CosminPerRam:
Closed this issue as it introduces many performance regressions, an alternative to this problem is #478.

PR #456, opened 2024-01-11
Add ID tests to CI

Douile:
Adds an action that runs when game IDs are changed and validates they match our current rules using rust-gamedig/gamedig-id-tests.


.github/workflows/id-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Use Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 18.x
+    - uses: Swatinem/rust-cache@v2
+      with:
+        cache-on-failure: 'true'
+    - run: cargo install --git https://github.com/gamedig/rust-gamedig.git gamedig-id-tests
+    - name: Run ID tests
+      run: node tools/run-id-tests.js
```
CosminPerRam, line 30: newlines pls :D

PR #459, opened 2024-01-11
chore: cleaner cli

CosminPerRam:
Attempts on making the CLI code a bit more cleaner.
Even with these changes I still think it doesn't look that well, open for suggestions.


CosminPerRam:
All seems to be good!

PR #461, opened 2024-01-12
chore: Cleaner and updated README

podrivo:
Hey! I've updated our README with some new info:

- Updated with Markdown Tables, for better visualization of each parameter
- Split into Required and Optional fields, for easier understanding
- Updated general descriptions with new info
- Added links for `GAMES_LIST.md` where needed
- Removed extra characters for smoother reading

This is 100% not needed, but I figure it could use an overall cleaner look.
Let me know your thoughts and I'll update everything accordingly! ✨

Thank you! (:

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,11 @@ capable of querying for the status of nearly any game or voice server.
 
 If a server makes its status publically available, GameDig can fetch it for you.
 
-Support is available on the [Discord server](https://discord.gg/NVCMn3tnxH) (for questions), or [GitHub Issues](https://github.com/gamedig/node-gamedig/issues) (for bugs).
+Support is available on the [Discord](https://discord.gg/NVCMn3tnxH) for questions, or [GitHub](https://github.com/gamedig/node-gamedig/issues) for bugs.
 
 ## Games List
-**node-GameDig** can query over 310 games (+ a few services)!  
-See the [GAMES_LIST.md](GAMES_LIST.md) file for the currently supported titles, not yet supported ones and notes about some of them.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,59 +28,50 @@ GameDig.query({
...
+| Key |   | Type | Description |
+|:---|:---|:---|:---|
+| **name** |   | string | Server name. |
+| **map** |   | string | Server map. |
+| **password** |   | boolean | If a password is required. |
+| **numplayers** |   | number | Number of players connected, via [A2S_INFO](https://developer.valvesoftware.com/wiki/Server_queries#A2S_INFO). |
+| **maxplayers** |   | number | Maximum number of connected players. |
+| **players** |   | array of objects | Note that this could be of a different length compared to **numplayers**. Via [A2S_PLAYER](https://developer.valvesoftware.com/wiki/Server_queries#A2S_PLAYER). |
+| **players** | **name** | string | If the player's name is unknown, the string will be empty. |
+| **players** | **raw** | object | Additional information about the player if available. |
```
CosminPerRam, line 66: I think this could be simpler to just do `players.name` rather than making another column for child properties.

podrivo: I agree! It was like that before, but I decided to complicate a bit... 😅

podrivo:
The `address` option was mentioned to be included in the documentation in this issue: https://github.com/gamedig/node-gamedig/issues/462, so might be a good idea to include it in this PR.

CosminPerRam (approved):
LGTM, thanks for being thoughtful regarding refactoring the README, the `address` documentation could be done in another PR cu keep things separated (:

Issue #438, opened 2023-12-18
feat: V-rising has an ofset of either 1 or 15


dgibbs64:
**What is this feature about?**
Doing a bit of research on port offset and found that V-rising servers tend to either have an offset of 1 or 15 depending upon the server provider. I am assuming that the developer's default is 15 but some providers choose to do a +1 for simplicity. So it might be a good idea to add a feature for multiple offsets.

**Additional context/references**
I am noticing this is a trend for many game servers with separate query ports so I believe allowing to query multiple common ports will increase the success rate. Maybe a default of checking the default port of set first then trying +1

CosminPerRam:
Following their [official server setup guide](https://github.com/StunlockStudios/vrising-dedicated-server-instructions) tells that the query port by default is +1 than the game port:
![image](https://github.com/gamedig/node-gamedig/assets/7972857/19d99a77-7342-487b-82df-2d4dd1e87d0d)

Have you found other guides or came across many servers with +15 offset?

podrivo:
I haven't found any docs about this either, but there's definitely a pattern on servers if you check this list: https://www.battlemetrics.com/servers/vrising

+1 and +15 on almost all servers.

CosminPerRam:
Oh, looks like it, alright.

podrivo:
Completed by https://github.com/gamedig/node-gamedig/pull/471 and https://github.com/gamedig/node-gamedig/pull/483.
It's ok to close! (:

PR #471, opened 2024-01-16
feat: make port_query_offset accept a number or an array of numbers

podrivo:
Hello!

This PR adds an alternative `port_query_offset`. I'm not sure if this all it needs to add an alternative `port_query_offset`, but it worked on the tests I did.

Currently with an option in the game `V Rising`. Related to this issue (https://github.com/gamedig/node-gamedig/issues/438).

### DIFF
--- a/lib/QueryRunner.js
+++ b/lib/QueryRunner.js
@@ -45,13 +45,9 @@ export default class QueryRunner {
       })
     }
 
-    let portOffsetArray = []
-    if (Array.isArray(gameQueryPortOffset)) {
-      portOffsetArray = gameQueryPortOffset
-    } else if (gameQueryPortOffset) {
-      portOffsetArray.push(gameQueryPortOffset)
-    } else {
-      portOffsetArray.push(0)
+    let portOffsetArray = gameQueryPortOffset
+    if (!Array.isArray(portOffsetArray)) {
+      gameQueryPortOffset ? portOffsetArray = [gameQueryPortOffset] : portOffsetArray = [0]
     }
 
     if (userOptions.port) {
--- a/lib/QueryRunner.js
+++ b/lib/QueryRunner.js
@@ -28,6 +28,7 @@ export default class QueryRunner {
     const {
       port_query: gameQueryPort,
       port_query_offset: gameQueryPortOffset,
+      port_query_offset_alt: gameQueryPortOffsetAlt,
       ...gameOptions
     } = lookup(userOptions.type)
     const attempts = []
@@ -47,7 +48,9 @@ export default class QueryRunner {
 
     if (userOptions.port) {
       if (!userOptions.givenPortOnly) {
-        if (gameQueryPortOffset) { addAttemptWithPort(userOptions.port + gameQueryPortOffset) }
+        if (gameQueryPortOffset) {addAttemptWithPort(userOptions.port + gameQueryPortOffset) }
+
+        if (gameQueryPortOffsetAlt) { addAttemptWithPort(userOptions.port + gameQueryPortOffsetAlt) }
 
         if (userOptions.port === gameOptions.port && gameQueryPort) { addAttemptWithPort(gameQueryPort) }
       }
--- a/lib/QueryRunner.js
+++ b/lib/QueryRunner.js
@@ -28,7 +28,6 @@ export default class QueryRunner {
     const {
       port_query: gameQueryPort,
       port_query_offset: gameQueryPortOffset,
-      port_query_offset_alt: gameQueryPortOffsetAlt,
       ...gameOptions
     } = lookup(userOptions.type)
     const attempts = []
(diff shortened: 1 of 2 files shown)


CosminPerRam:
Heyo, thanks for taking care of this.
I have mixed feelings regarding adding another field option for this -- what if we want to have 3 offsets?
I think making `gameQueryPortOffset` an array would be a better choice, but as the majority of games have only a single offset, it would be great if we could handle this as a `number` or as an `array of numbers`.

lib/QueryRunner.js
```diff
@@ -47,16 +47,34 @@ export default class QueryRunner {
 
     if (userOptions.port) {
       if (!userOptions.givenPortOnly) {
-        if (gameQueryPortOffset) { addAttemptWithPort(userOptions.port + gameQueryPortOffset) }
+
```
CosminPerRam, line 50: Please remove the stray newlines.

lib/games.js
```diff
@@ -2656,7 +2656,7 @@ export const games = {
     name: 'V Rising',
     options: {
       port: 27015,
-      port_query_offset: 1,
+      port_query_offset: [1, 15],
```
CosminPerRam, line 2659: This should be done in a separate PR to link to that issue.

lib/QueryRunner.js
```diff
@@ -47,16 +47,34 @@ export default class QueryRunner {
...
+        if (gameQueryPortOffset) {
+          if (Array.isArray(gameQueryPortOffset)) {
```
CosminPerRam, line 52: The repeating if's could be resolved be preprocessing `gameQueryPortOffset`.
Before the first if, check if it is an array, if not, make one by the sole value (if any).

Afterwards, its just a matter of `.forEach` it and also the `if (gameQueryPortOffset)` will not be required anymore.

CosminPerRam (changes requested):

lib/QueryRunner.js
podrivo: Removed!

lib/QueryRunner.js
podrivo: I think I understood what you meant. Thank you! 🙌 
Just added a new code. Let me know!

lib/QueryRunner.js
```diff
@@ -45,18 +45,25 @@ export default class QueryRunner {
       })
     }
 
+    let portOffsetArray = []
+    if (Array.isArray(gameQueryPortOffset)) {
+      portOffsetArray = gameQueryPortOffset
+    } else if (gameQueryPortOffset) {
+      portOffsetArray.push(gameQueryPortOffset)
+    } else {
+      portOffsetArray.push(0)
```
CosminPerRam, line 54: This is not needed.

lib/QueryRunner.js
```diff
@@ -45,18 +45,25 @@ export default class QueryRunner {
       })
     }
 
+    let portOffsetArray = []
+    if (Array.isArray(gameQueryPortOffset)) {
+      portOffsetArray = gameQueryPortOffset
+    } else if (gameQueryPortOffset) {
+      portOffsetArray.push(gameQueryPortOffset)
```
CosminPerRam, line 52: Even simpler would be:
```js
let portOffsetArray = gameQueryPortOffset
if (!Array.isArray(portOffsetArray)) { // the given offset is not an array apparently, it's a single value, so lets fix that
  portOffsetArray = [gameQueryPortOffset]
}
```
Sorry if I seem to be a little petty about this, just thought it could be better (:

podrivo: Please, by all means! Your input is very much appreciated!

podrivo: I did some testing and when we don't have `port_query_offset` present, we get `gameQueryPortOffset` set as `undefined` and this will mess with the attempts for when we don't have a port set by the user.

```js
} else if (gameOptions.port) {
      portOffsetArray.forEach((portOffset) => { addAttemptWithPort(gameOptions.port + portOffset) })
}
```

The line `portOffsetArray.push(0)` was basically setting a `[0]` to the `gameQueryPortOffset`.
So if I remove `port_query_offset` from `games.js` and run this below, I'll get an error.

`node bin/gamedig --type vrising 45.137.247.66`

podrivo: I've updated to the code below, so we make sure to always have a `0` to be used when no port is provided.
Let me know if it makes sense! (: 

```js
let portOffsetArray = gameQueryPortOffset
if (!Array.isArray(portOffsetArray)) {
  gameQueryPortOffset ? portOffsetArray = [gameQueryPortOffset] : portOffsetArray = [0]
}
```

CosminPerRam: great catch, thanks.

CosminPerRam (approved):

Issue #464, opened 2024-01-14
feat: Call of Duty Black Ops 3


dgibbs64:
**What is this feature about?**
Add Call of Duty Black Ops 3.

**Additional context/references**
 It has the same query type as Modern Warfare 3

podrivo:
Hey @dgibbs64.

Thank you! We'll make sure to add this as an option in the games list.
In the mean time, you can currently use the Modern Warfare 3 query type to fetch server info:

```js
gamedig --type codmw3 --pretty 108.61.158.93:27017

{
  "name": "Aperture BO3 Classic Svr",
  "map": "mp_stronghold",
  "password": false,
  "raw": {
    "protocol": 17,
    "folder": "usermaps",
    "game": "Black Ops 3",
    "appId": 311210,
    "numplayers": 0,
    "numbots": 0,
    "listentype": "d",
    "environment": "w",
    "secure": 1,
    "version": "1.0.0.0",
    "steamid": "90179601157573656",
    "tags": [
      "\\gametype\\tdm\\dedicated\\true\\ranked\\false\\hardcore\\false\\zombies\\false\\modName\\usermaps\\playerCount\\0"
    ]
  },
  "maxplayers": 18,
  "players": [],
  "bots": [],
  "connect": "108.61.158.93:27017",
  "ping": 0
}
```

CosminPerRam:
Closed by #473.

PR #473, opened 2024-01-16
feat: Add Call of Duty: Black Ops 3 support

podrivo:
It closes this issue (https://github.com/gamedig/node-gamedig/issues/464).

Also, I'm not sure about the `port_query_offset`, but it worked on the tests I did.


GAMES_LIST.md
```diff
@@ -61,6 +61,7 @@
 | cod4mw               | Call of Duty 4: Modern Warfare                   |                                             |
 | codenamecure         | Codename CURE                                    | [Valve Protocol](#valve)                    |
 | codenameeagle        | Codename Eagle                                   |                                             |
+| codbo3               | Call of Duty: Call of Duty: Black Ops 3          |                                             |
```
CosminPerRam, line 64: Repeated "Call of Duty:"

lib/games.js
```diff
@@ -458,6 +458,14 @@ export const games = {
     },
     release_year: 2007
   },
+  codbo3: {
+    name: 'Call of Duty: Black Ops 3',
+    release_year: 2015,
+    options: {
+      port_query_offset: 2,
```
CosminPerRam, line 465: If we have a `port_query_offset`, we'd want to also have `port` or `port_query` (to be explicit about from what we are adding/substracting without searching for the default protocol port).
I've found a guide that lists `27017` as default port and all of the 3 servers I found online were on that port, and the query port seems to be the same, so just `port: 27017` is required here.

CosminPerRam (changes requested):
Thank you! Just some small changes please (:

GAMES_LIST.md
podrivo: Oops! Sorry about that. Fixed!

lib/games.js
podrivo: That's good to know! And thank you for founding the default port.
I've copied this from the `Call of Duty: Modern Warfare 3`, so I'll make sure to take a look at as well:

https://github.com/gamedig/node-gamedig/blob/b481425dbf1d50dcb3e756897d7cc15a46cc04c9/lib/games.js#L477-L484

CosminPerRam (approved):
Thanks!

CosminPerRam:
Oops, seems that this need to be rebased.

CosminPerRam (approved):

PR #475, opened 2024-01-16
feat(tools/attempt): test against all protocols

podrivo:
Pull protocol list from `/protocols/index.js` and adds to `/tools/attempt_protocols.js`

### DIFF
--- a/tools/attempt_protocols.js
+++ b/tools/attempt_protocols.js
@@ -16,11 +16,8 @@ if (argv._.length >= 1) {
 
 const gamedig = new GameDig(options)
 
-let protocolList = []
-Object.keys(protocols).forEach((key) => protocolList.push(key))
-
 const run = async () => {
-  for (const protocol of protocolList) {
+  for (const protocol of Object.keys(protocols)) {
     try {
       const response = await gamedig.query({
         ...options,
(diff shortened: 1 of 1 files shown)


tools/attempt_protocols.js
```diff
@@ -15,10 +16,11 @@ if (argv._.length >= 1) {
 
 const gamedig = new GameDig(options)
 
-const protocols = ['valve', 'gamespy1', 'gamespy2', 'gamespy3', 'goldsrc', 'minecraft', 'quake1', 'quake2', 'quake3', 'unreal2', 'valve']
+let protocolList = []
+Object.keys(protocols).forEach((key) => protocolList.push(key))
 
 const run = async () => {
-  for (const protocol of protocols) {
+  for (const protocol of protocolList) {
```
CosminPerRam, line 23: I think its the same thing to do `for (const protocol of Object.keys(protocols))`, could you check?

podrivo: Nice! Indeed it is! Thank you! ✨

podrivo:
I've noticed during testing, that the `beammpmaster` always respond to something, regardless of the IP.
Is this how it's supposed to be?

CosminPerRam:
Ahm, totally forgot about master servers.
It is supposed to always respond, but in our use case we dont want it to.
I was thinking on separating game protocols and services (Discord, BeamMP master etc), but that'll take a considerable amount of work, so we're not doing this now.
You could revert to your initial solution, but filter out these protocols.

podrivo:
Nice! I've updated the code to filter out these services!

CosminPerRam (approved):

PR #478, opened 2024-01-16
feat: port caching

CosminPerRam:
Adds the option (enabled by default) to cache ports.

This is the most beneficial when you query the same server (identified by resolved address and port) and it doesn't end up successful on the first attempted port (e.g. specify the query port without `--givenPortOnly` or games that might have multiple offsets (soon to come)).

An example of this would be:
Query `DayZ` server A (with 1 maximum attempt), where we know the query port of that specific server (P1):
1st attempt: specified port + offset (P2)... failure.
2nd attempt: specified port (P1)... success. -- add this port as cached.

Some time later, query the same server (A) with the same port (P1):
1st attempt: cached port... success.

It would only regress if the server changes the query port whilst your service is running, which is not very likely but the other edge cases heavily outweigh this.

This replaces #454, that PR introduces backwards-compatibility regressions by always trying the specified user port firstly, as many users of [Homepage](https://github.com/gethomepage/homepage) or [Uptime Kuma](https://github.com/louislam/uptime-kuma) just put in their default connect port. The attempt order ends up with the first attempted port to be the specified port and the offset, the PR would change this behaviour by making the user given port the first one, making these queries slower with the extra attempt.

### DIFF
--- a/bin/gamedig.js
+++ b/bin/gamedig.js
@@ -6,10 +6,11 @@ import Minimist from 'minimist'
 import { GameDig } from './../lib/index.js'
 
 const argv = Minimist(process.argv.slice(2), {
-  boolean: ['pretty', 'debug', 'givenPortOnly', 'requestRules', 'requestRulesRequired', 'requestPlayersRequired', 'stripColors'],
+  boolean: ['pretty', 'debug', 'givenPortOnly', 'requestRules', 'requestRulesRequired', 'requestPlayersRequired', 'stripColors', 'portCache'],
   string: ['guildId', 'listenUdpPort', 'ipFamily'],
   default: {
-    stripColors: true
+    stripColors: true,
+    portCache: true
   }
 })
 
--- a/lib/QueryRunner.js
+++ b/lib/QueryRunner.js
@@ -84,7 +84,9 @@ export default class QueryRunner {
 
         try {
           const response = await this._attempt(attempt)
-          this.portCache[`${userOptions.address}:${userOptions.port}`] = attempt.port
+          if (attempt.portCache) {
+            this.portCache[`${userOptions.address}:${userOptions.port}`] = attempt.port
+          }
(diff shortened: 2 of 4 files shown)


lib/QueryRunner.js
```diff
@@ -45,6 +47,12 @@ export default class QueryRunner {
       })
     }
 
+    const cachedPort = this.portCache[`${userOptions.address}:${userOptions.port}`]
+
+    if (cachedPort && optionsCollection.portCache) {
+      addAttemptWithPort(cachedPort)
```
Douile, line 53: If the cached port fails this will result in 2 attempts to the same port.

Douile (commented):
Overall I think this is a great idea. I have a few implementation nitpicks though.

lib/QueryRunner.js
CosminPerRam: In the case whereas the server query port changes during runtime, right?

Douile: It depends on the retry options, but would only occur if there was a timeout or other failure on the cached port yes.

bin/gamedig.js
```diff
@@ -6,10 +6,11 @@ import Minimist from 'minimist'
 import { GameDig } from './../lib/index.js'
 
 const argv = Minimist(process.argv.slice(2), {
-  boolean: ['pretty', 'debug', 'givenPortOnly', 'requestRules', 'requestRulesRequired', 'requestPlayersRequired', 'stripColors'],
+  boolean: ['pretty', 'debug', 'givenPortOnly', 'requestRules', 'requestRulesRequired', 'requestPlayersRequired', 'stripColors', 'portCache'],
```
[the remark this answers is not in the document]
CosminPerRam: I'll leave it like this for now, the CLI does need a wider overhaul and this could be part of that.

PR #485, opened 2024-01-17
chore: rename maxAttempts option to maxRetries

CosminPerRam:
Renames the `maxAttempts` option to `maxRetries`.
An attempt is defined by its query settings (usually consisting of different specified ports), a retry is when you do an attempt again.
(replacing the word `retry` with `attempt` in this phrase points out why I thought this could be renamed)


README.md
```diff
@@ -41,7 +41,7 @@ Confused on how this works, or you want to see more? Checkout the [examples](/ex
...
 | **address**                | string  | undefined | Override the IP address of the server skipping DNS resolution. When set, host will not be resolved, instead address will be connected to. However, some protocols still use host for other reasons e.g. as part of the query. |
 | **port**                   | number  | Game port | Connection port or query port for the game server. Some games utilize a separate "query" port. If specifying the game port does not seem to work as expected, passing in this query port may work instead.                    |
-| **maxAttempts**            | number  | 1         | Number of attempts to query server in case of failure.                                                                                                                                                                        |
+| **maxRetries**             | number  | 1         | Number of retries to query server in case of failure. Note that this amount applies multiplies with the number of attempts.                                                                                                   |
```
podrivo, line 44: Maybe a typo here?

podrivo:
Shorter and easier to write as well 😅

cetteup:
Note that this is a breaking change to the public interface, so this should be released with a major version bump (at least when following semantic versioning).

podrivo:
Maybe we should consider having both in the code for backwards compatibility?

CosminPerRam:
It is indeed a breaking change, thus intended for v5. Forgot to mention this, sorry.
> Maybe we should consider having both in the code for backwards compatibility?

For v5, no, as it already has many changes (and I will look into a CLI overhaul).
For v4 (backporting), I wouldn't really see a reason to backport this as its just a rename.

PR #486, opened 2024-01-17
feat: breadth attempt order

### LINKED ISSUE #479: Change retry logic to "breadth first"
**What is this feature about?**
Currently we handle retries by queuing up a list of possible query settings then iterating this list and trying each one for the amount of retries attempts before moving on to the next set of possible settings.

This looks like this (max retries 2 with 3 different settings):
```
Attempt 1: settings=0 retries=0
Attempt 2: settings=0 retries=1
Attempt 3: settings=1 retries=0
Attempt 4: settings=1 retries=1
Attempt 5: settings=2 retries=0
Attempt 6: settings=2 retries=1
```

Depending on the situation (how good first guess is) it might be better to move on to the next possible query settings after the previous fails. This would look something like this:

```
Attempt 1: settings=0 retries=0
Attempt 2: settings=1 retries=0
Attempt 3: settings=2 retries=0
Attempt 4: settings=0 retries=1
Attempt 5: settings=1 retries=1
Attempt 6: settings=2 retries=1
```

By iterating this way we explore the possible ports the server could be on quicker (when maxRetries is > 1) so in theory we would return a result faster when the correct port is not the first in the list.

This would be slower in the case that there is a timeout querying the first guess but the first guess is the correct port.

**Additional context/references**
https://github.com/gamedig/node-gamedig/blob/b481425dbf1d50dcb3e756897d7cc15a46cc04c9/lib/QueryRunner.js#L65-L88

CosminPerRam:
Closes #479.
Also adds an option to disable this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Confused on how this works, or you want to see more? Checkout the [examples](/ex
 | **requestPlayersRequired** | boolean | false     | Valve games only. Querying players is always required to have a response or the query will timeout. Some [games](GAMES_LIST.md) may not provide a players response.                                                           |
 | **stripColors**            | boolean | true      | Enables stripping colors for protocols: unreal2, savage2, quake3, nadeo, gamespy2, doom3, armagetron.                                                                                                                         |
 | **portCache**              | boolean | true      | After you queried a server, the second time you query that exact server (identified by specified ip and port), first add an attempt to query with the last successful port.                                                   |
+| **noBreadthOrder**         | boolean | false     | Enable the behaviour of retrying an attempt X times followed by the next attempt X times, otherwise try attempt A, then B, then A, then B until reaching the X retry count of each.                                           |
 
 ## Query Response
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -52,6 +52,7 @@ Confused on how this works, or you want to see more? Checkout the [examples](/ex
 | **requestPlayersRequired** | boolean | false     | Valve games only. Querying players is always required to have a response or the query will timeout. Some [games](GAMES_LIST.md) may not provide a players response.                                                           |
 | **stripColors**            | boolean | true      | Enables stripping colors for protocols: unreal2, savage2, quake3, nadeo, gamespy2, doom3, armagetron.                                                                                                                         |
 | **portCache**              | boolean | true      | After you queried a server, the second time you query that exact server (identified by specified ip and port), first add an attempt to query with the last successful port.                                                   |
+| **noBreadthOrder**         | boolean | false     | Enable the behaviour of retrying an attempt X times followed by the next attempt X times, otherwise try attempt A, then B, then A, then B until reaching the X retry count of each.                                           |
```
podrivo, line 55: I'm not sure I follow the description. Is it working alongside `maxRetries`?

podrivo: Nice! I've re-read it and it's clearer now. Thanks! (:

Issue #465, opened 2024-01-14
bug: Hidden and Dangerious2 hiddendangerous2 Invalid game


dgibbs64:
**Describe the bug**
`{"error":"Invalid game: hiddendangerous2"}`

**Steps To Reproduce**
server I found 78.47.255.224:11005 attempte to query it with hiddendangerous2
`gamedig --type hiddendangerous2 78.47.255.224:11005`
**Expected behavior**
Game is a listed and should work

podrivo:
Hello @dgibbs64! This was recently updated and not yet published.
It's currently named as `hiddendangerous2`, but you should use `had2` for now.

```js
gamedig --type had2 --pretty 78.47.255.224:11005

{
  "name": "=RpR= Justice",
  "map": "Nm2(Frontline)",
  "password": false,
  "raw": {
    "hostname": "=RpR= Justice",
    "gamever": 1.12,
    "hostport": 11005,
    "mapname": "Nm2(Frontline)",
    "numplayers": 1,
    "maxplayers": 12,
    "gamemode": "0",
    "gametype": "Cooperative",
    "isdedicated": 1,
    "password": 0,
    "voicechat": 0,
    "expansion": 2,
    "team_t0": "axis",
    "team_t1": "allied",
    "score_t0": 0,
    "score_t1": 0,
    "teams": {}
  },
  "maxplayers": 12,
  "players": [
    {
      "name": " ??? ???????",
      "raw": {
        "score": 0,
        "deaths": 0,
        "ping": 0,
        "team": 1
      }
    }
  ],
  "bots": [],
  "connect": "78.47.255.224:11005",
  "ping": 244
}
```

CosminPerRam:
Just as @podrivo mentioned, that id is for v5 and `had2` is what its on v4.

Issue #374, opened 2023-10-07
Update game ids to follow a standard


CosminPerRam:
## What are game ids
Game ids are important lowercase alphanumeric string that we use when specifying a Game.
Currently, there is no standard on how an ID is made. 
In #371 I said "lets dont make 2-characters ids anymore" because I didn't thought that id represented the game name, and in other pr's for adding games support, the author decided the game id however they wanted (because there are no rules for this).

## What could be done
This issue suggests applying [rust-GameDig](https://github.com/gamedig/rust-gamedig)'s id system added in the [contributing](https://github.com/gamedig/rust-gamedig/pull/108) pr.

## Problems
Although it is stated (at the time of making this issue) in the rust-GameDig's readme that "Warning: This project goes through frequent API breaking changes [...]" this is not done so here, the new ID system will break a bunch of ids and although the next release is a major one, breaking changes could be done, but are not preferred.

One solution for this would be to keep the old ids and add the new ones (until another major version comes), but I haven't verified yet if there are any new-old colliding ones (for example, a new game's id would be `ase` but there is already an old id `ase`, so we cant have both), if this would be possible, I guess its the way to go.

This issue keeps track on deciding what to do about this manner.

dgibbs64:
Totally agree with this. Setting a standardised naming convention and documenting it will be massively helpful. To reduce possible conflict longer names might be best e.g `americasarmy3` however you also run into issues such as `csgo` which would be too long e.g `counterstrikeglobaloffensive`.

with LinuxGSM I use shortnames https://github.com/GameServerManagers/LinuxGSM/blob/master/lgsm/data/serverlist.csv and havnt had too many issues. The longest shortname used is 8 characters.

podrivo:
Was this issue addressed with this pull (https://github.com/gamedig/node-gamedig/pull/415)?

CosminPerRam:
Not entirely, there are a few that fails (see CI) and I was thinking on providing a 'translation layer' between the old and new ids (#435 helps a lot regarding this, but I just didn't had the time lately).

podrivo:
PR with new ids and translation layer merged https://github.com/gamedig/node-gamedig/pull/496
ID testing issue over at https://github.com/gamedig/rust-gamedig/issues/184

CosminPerRam:
Fully compliant as per [331d7bb](https://github.com/gamedig/node-gamedig/commit/331d7bbc538df0b87971ab8a797c41cf98d2e48a).

PR #487, opened 2024-01-18
feat: Add game type alias

podrivo:
This adds an `alias` for GIDs, so it will work for the official current supported GID and also previous or new GIDs.
Got this idea from this issue https://github.com/gamedig/node-gamedig/issues/465.

Noticed that **Hidden and Dangerious2** had the GID updated, but also a few others had the GID updated as well (see PR https://github.com/gamedig/node-gamedig/pull/415), such as **Age of Chivalry**, **Action: Source**, **Ark: Survival Evolved**, **Left 4 Dead** etc. This will enable backwards compatibility for those previous GIDs.

Also, should I include alias GIDs on `lib/games.js` and update `GAMES_LIST.md` on this PR or should I open another?

Let me know if this makes sense. Thank you! (:

—

Example:
```js
hiddendangerous2: {
  name: 'Hidden & Dangerous 2',
  alias: 'had2',
  release_year: 2003,
  options: {
    port: 11001,
    port_query_offset: 3,
    protocol: 'gamespy1'
  }
}
```


CosminPerRam:
This is definitely something that has to be done.
> Also, should I include alias GIDs on lib/games.js and update GAMES_LIST.md on this PR or should I open another?

Would advocate to do so here, so we are sure we cover every possible problem that could appear while doing so.

I wouldn't name this field as `alias`, I for one think `oldId` would be more appropriate to what it really is.

While adding this 'translation' layer, you could find cases where:
- Newly supported Game has id X
- Older supported Game with new id is Y but old id is X

And what would we want to do here? Prioritize newly ids or older ones? This'll go multiple ways:
1. Look only for new ids.
2. Prioritize new ids, if none is found, look for the old ones.
3. Prioritize old ids, if none is found, look for the newer ones.
4. Look only for old ids.

We'll have to decide the default behaviour, also this 'translation' layer could (and I guess will) be removed after a new major version (v6).

podrivo:
Nice! I'll start adding GIDs from previous commits to this PR, so we keep everything together! (:

—

About the name, I'm not strongly attached to `alias`, but I would argue that this is not time based, as in new/old, but only as a different name format. More as like how users would use and less like how we maintain it.

Almost like having a full and a short name. So for example: Counter Strike 2, the full main format would be `counterstrike2` and the short format would be `cs2`.  Users would probably default to use `cs2` as it's shorter and easier to remember, but for us maintaining it, `counterstrike2` would make more sense and fit more inside [our guidelines](https://github.com/gamedig/node-gamedig/blob/master/CONTRIBUTING.md#naming) (which we should also make sure it's up to date to the new commits).

This would exempt us from thinking about new/old and prioritization when looking for GIDs, and it would work in all major versions, even if we decide to change our naming guidelines.

Let me know what you think!

podrivo:
I'm late to the party overall, and so I'm just putting everything together and checking issues and pulls around.
So I was checking this https://github.com/gamedig/node-gamedig/issues/374 and this https://github.com/gamedig/node-gamedig/pull/435/.

Will consider these links and think this through!

CosminPerRam:
> Let me know what you think!

I agree that `cs2` is nicer to remember and type rather than `counterstrike2`, but, as stated partially in #415:
At the end of the day the ids dont matter that much, as if one service wants to utilize this library and would let the user choose the game that they wanna query, they wont let them choose ids, they'll choose names (why would they choose ids?), so what exactly is the id it doesn't really matters here.
Could be totally wrong about this.

As for having a 'short name', it could be done, but I don't really see much value in having this and maintaining it.

podrivo:
Considering everything, I think this is just a minor detail in the big scheme of games ids. I do like the idea of being able to use `cs2` as well as `counterstrike2`. There's just something that bugs me having `csgo`, `css` but `counterstrike2`. It feels off haha. And I think for some users it will be more intuitive having the social/cultural name. I'm sure there's more examples.

But for now, let's focus on the other open discussions, and when it's time we can talk more about it. I don't want to bring even more things to consider, so I'll close this for now, as it's all linked and we can reference later.

Thank you! (:

podrivo:
It would be good to have more insight on how to proceed about this implementation! (:

a-sync:
Suggestion: merge the old id check and the alias check into `checkAliases` for example. Keep a single field in the games list like `extra.aliases` and make it a (string or) string array.  

Could have both old ids and aliases and would make it extendable.

podrivo:
@a-sync Thanks for joining! ✨ 

That's the original idea, but @CosminPerRam has a different view about how to deal with this.
I was waiting for his direction on how to proceed.

CosminPerRam:
> unless there are conflicting old/new IDs i don't see a reason why not just check for all possible values and loose the option, its not that big of an overhead imo

It indeed isn't a big overhead, but we cant guarantee that eventually we wouldn't be adding a new id that conflicts with the old ones, thus breaking old behaviour, so I think its better to separate them so that the users of the lib are aware that they will eventually be removed.

podrivo:
It's been more than 2 years, it has conflicts to resolve and no other comments suggesting this will move forward, so closing this PR.

Issue #491, opened 2024-01-19
feat: Support Palworld game


Maxence1502:
**What is this feature about?**

Hello,

Is it possible to support this game : https://store.steampowered.com/app/1623730/Palworld/

Thank you :)

CosminPerRam:
I couldn't find any info at all regarding direct server queries.
Seems that the game is available through the steam servers browser:
![image](https://github.com/gamedig/node-gamedig/assets/7972857/7d1d1c83-e980-4631-9feb-49d0744ff87d)
(screenshot courtesy of @jammsen)
And its seems the game has the port `27015` opened (which usually indicates a valve protocol, but it doesnt respond on it): 
![image](https://github.com/gamedig/node-gamedig/assets/7972857/5a1295e4-2edf-4a29-b128-ebf2a697eb1c)

jonathanprl:
It's using EOS and works similar to the game `asa`, although requires extra EOS authentication steps. I've got something working although it's not querying every Palworld server, but does query some...

https://github.com/jonathanprl/node-gamedig/commit/514101ff81a80b6f1ffc85b80ccbd8d9ce055602

CosminPerRam:
Forgot to mention that the game's engine is unreal so it probably runs EOS.

Well, a partial solution is better than no solutions.
> although requires extra EOS authentication steps

Hmm, that could go the same about some other EOS games (cc as I know you've looked into other games: @GuilhermeWerner).

> https://github.com/jonathanprl/node-gamedig/commit/514101ff81a80b6f1ffc85b80ccbd8d9ce055602

Feel free to make a PR whenever you think it's meargeable (:

jammsen:
Maybe that helps out a little
https://tech.palworldgame.com/community-server-guide

jammsen:
Feel free to use my server as test-server too:
https://www.battlemetrics.com/servers/palworld/25160493

GuilhermeWerner:
> Forgot to mention that the game's engine is unreal so it most probably runs EOS.
> 
> Well, a partial solution is better than no solutions.
> 
> > although requires extra EOS authentication steps
> 
> Hmm, that could go the same about some other EOS games (cc as I know you've looked into other games: @GuilhermeWerner).
> 
> > [jonathanprl@514101f](https://github.com/jonathanprl/node-gamedig/commit/514101ff81a80b6f1ffc85b80ccbd8d9ce055602)
> 
> Feel free to make a PR whenever you think it's meargeable (:

[...]

PR #495, opened 2024-01-20
feat: add support for Palworld

jonathanprl:
Worth noting that this will only query game servers that are listed in EOS matchmaking. I've noticed at the moment that Palworld, as it's currently booming in popularity, does not include all game servers on it's matchmaking list.

You can also see currently when you host a dedicated server, that the "create session" requests that are made to EOS often fail. When this happens, the game server does not appear in EOS and you can only directly connect in-game.

As far as I'm aware there's not currently a way to directly query a Palworld server.

### DIFF
--- a/protocols/epic.js
+++ b/protocols/epic.js
@@ -15,7 +15,8 @@ export default class Epic extends Core {
     this.clientSecret = null
     this.deploymentId = null
     this.epicApi = 'https://api.epicgames.dev'
-    this.clientAccessToken = null
+    this.alternativeAuth = false // Some games require a client access token to POST to the matchmaking endpoint.
+
     this.deviceIdAccessToken = null
     this.accessToken = null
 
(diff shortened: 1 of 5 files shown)


jonathanprl:
Worth noting that this will only query game servers that are listed in EOS matchmaking. I've noticed at the moment that Palworld, as it's currently booming in popularity, does not include all game servers on it's matchmaking list.

You can also see currently when you host a dedicated server, that the "create session" requests that are made to EOS fail.

As far as I'm aware there's not currently a way to _directly_ query a Palworld server.

CosminPerRam:
Thanks for the PR, a partial solution is better than no solution.

> You can also see currently when you host a dedicated server, that the "create session" requests that are made to EOS often fail. When this happens, the game server does not appear in EOS and you can only directly connect in-game.

Don't quote me on this, but I've heard around (cant pin-point a source) that EOS currently has issues because of the amount of in game sessions and stuff related to it.

GuilhermeWerner:
Well, this implementation works on @jammsen server. And on other random servers I tested.

protocols/epic.js
```diff
@@ -15,20 +15,28 @@ export default class Epic extends Core {
     this.clientSecret = null
     this.deploymentId = null
     this.epicApi = 'https://api.epicgames.dev'
+    this.alternativeAuth = false // Some games require a client access token to POST to the matchmaking endpoint.
```
CosminPerRam, line 18: `alternativeAuth` doesnt sounds very indicative of what it is, this could be renamed to something like `authByExternalToken` (or something else).

CosminPerRam (changes requested):

CosminPerRam:
This PR closes #491.

CosminPerRam:
Forgot to mention in my review to also add the game in changelog and games_list, I've done it in the previously referenced commit, sorry!

Issue #493, opened 2024-01-20
feat: GIDs translation layer between v4 and v5?


CosminPerRam:
**What is this feature about?**
PR #415 introduced the biggest breaking change in the next major version, a standard of creating game ids (GIDs).
This is a problem as user cannot migrate easily to this version without rechecking what their game's new id is.

While adding this 'translation' layer, we could find cases where:
* Newly supported Game has id X
* Older supported Game with new id is Y but old id is X

And what would we want to do here? Prioritize newly ids or older ones? 
This could go multiple ways:
1. Look only for new ids.
2. Prioritize new ids, if none is found, look for the old ones.
3. Prioritize old ids, if none is found, look for the newer ones.
4. Look only for old ids.

I for one think that we should look only for new ids by default (solution 1) and have an additional field that does solution 3.

**Additional context/references**
Talked some more about this in #487.

podrivo:
Maybe instead of having a layer of 'translation' built into the code, should we consider having a separate markdown file with this 'translation' from v4 to v5? A migration document.

Something like a table of GIDs:

| v4 |  | v5
|:---|:---|:---
| cs2 | → | counterstrike2
| had2 | → | hiddendangerous2
| left4dead2 | → | l4d2

CosminPerRam:
A 'migration guide' for v4->v5 is also something that I thought about and your suggestion is really good, thanks.

The problem in the grand scheme of things are projects such as Homepage or Uptime Kuma: 
If they'll want to migrate from v4 to v5, they cannot do this because it'll break all the current user-set GameDig gids, an option here is to add a 'banner' on their side to indicate this or introduce their own translation layer, but this is really our 'fault', so we should resolve it as smoothly as possible for them.

podrivo:
I'm unfamiliar with Homepage or Uptime Kuma. How they work and why can't they migrate versions?

CosminPerRam:
Let's take a look at the support for GameDig on HomePage (link [here](https://gethomepage.dev/latest/widgets/services/gamedig/)):
![image](https://github.com/gamedig/node-gamedig/assets/7972857/9de34590-33f2-43ea-a675-ddee17394daf)
Now let's suppose the 'type' value was `cs2` (to demonstrate this):
The latest homepage version is using gamedig version `4.3.0`, in that version the GID for `Counter Strike 2` is `cs2`, but in v5 it will be `counterstrike2`.
[...]

PR #496, opened 2024-01-21
feat: add checking for older GIDs

podrivo:
This PR adds a new property `games.extra.old_id`, that is being used to add older game type IDs into the `game-resolver.js`. This makes sure that the new naming convention doesn't brake third party products that rely on gamedig.

It has a similar implementation as this PR (https://github.com/gamedig/node-gamedig/pull/487), but has a different essence on closing the gap about new/old GIDs, as discussed on https://github.com/gamedig/node-gamedig/issues/493 and https://github.com/gamedig/node-gamedig/issues/374.

Adds the option to skip this check:

- `skipOldIDs` (defaults to `false`). Query will skip checking for older game type IDs.

—

I've included a tool for finding duplicates, that checks for new and old ids, and can be used in the CI. I'm not sure how to add this in the CI or if the code will work in the CI. Let me know and I'll update it!

```
tools/find_id_duplicates.js
```

—

Updates `README.md`, ~the Naming section on `CONTRIBUTING.md` and tweaks details on `GAMES_LIST.md`~.
Adds a new file `MIGRATION.md` with a list of IDs from v4 to v5. The list was generated using `tools/find-id-changes.js`


podrivo:
I've included a tool for finding duplicates, that can be used in the CI testing. It checks for new and old ids.

podrivo:
CI ID tests is failing. Should we take a look at that as well with this PR and make sure everything is aligned?
I'm not sure how to update this... Feels like it needs env vars and the rust repo. 

Here's the error log:

```
IDFail {
thread 'main' panicked at crates/id-tests/src/main.rs:30:5:
assertion failed: failed.is_empty()
    game_id: "cs2d",
    game_name: "CS2D",
    expected_id: "c2d",
    rule_stack: [
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
        NumbersAreTheirOwnWord,
        MoreThanTwoWordsMakeAcronym,
    ],
}
IDFail {
    game_id: "eldewrito",
    game_name: "Halo Online ",
    expected_id: "haloonline",
    rule_stack: [
        TwoWordsOrLessUseFullWords,
    ],
}
IDFail {
    game_id: "il2sturmovik",
    game_name: "2 Sturmovik",
    expected_id: "twosturmovik",
    rule_stack: [
        IfModForQueriesProcessOnlyModName,
        IfFirstWordNumberNoDigits,
        TwoWordsOrLessUseFullWords,
    ],
}
IDFail {
    game_id: "il2sturmovik",
    game_name: "IL-2 Sturmovik",
    expected_id: "i2s",
    rule_stack: [
        MoreThanTwoWordsMakeAcronym,
    ],
}
IDFail {
    game_id: "terrariatshosck",
    game_name: " TShock",
    expected_id: "tshock",
    rule_stack: [
        IfModForQueriesProcessOnlyModName,
        TwoWordsOrLessUseFullWords,
    ],
}
IDFail {
    game_id: "terrariatshosck",
    game_name: "Terraria - TShock",
    expected_id: "terrariatshock",
    rule_stack: [
        TwoWordsOrLessUseFullWords,
    ],
}
IDFail {
    game_id: "jb007n",
    game_name: "James Bond 007: Nightfire",
    expected_id: "jb0n",
    rule_stack: [
        MoreThanTwoWordsMakeAcronym,
    ],
}
7 (2%) IDs didn't match naming rules
```

lib/game-resolver.js
```diff
@@ -9,7 +11,13 @@ export const lookup = (type) => {
...
   }
 
-  const game = games[type]
+  let game = games[type]
+  
+  Object.keys(games).forEach((id) => {
+    if (games[id].extra && games[id].extra.old_id && !options.skipOldIDs) {
+      game = games[id]
+    }
+  })
```
CosminPerRam, line 20: This does the `skipOldIDs` check on every game, which is not ideal.
You should wrap this iteration in that `skipOldIDs` check (so it checks just one time).
Also, instead of `games[id].extra && games[id].extra.old_id` I think you could do `games[id]?.extra.old_id`

tools/generate_games_list.js
```diff
@@ -21,12 +21,17 @@ sortedGamesIds.forEach(key => {
 })
 
 let generated = ''
-generated += '| GameDig Type ID | Name | See Also\n'
-generated += '|---|---|---\n'
+generated += '| GameDig Type ID | Old ID | Name | See Also\n'
```
CosminPerRam, line 24: Hmmm, i think we shouldn't add the Old ID as a column in the `GAMES_LIST` file.
These ids will eventually be removed in a future version and I think by keeping them would incentivese users to keep them or for new users to use the olds ids as they would appear to be more pleasing (related to the aliasing idea you gave). Thoughts?

Having a markdown file dedicated to migrating from v4 to v5 would be ideal and these `old vs new` ids would be perfect there.

tools/find_id_duplicates.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  return false
+}
+
+if (hasDuplicates(ids)) {
+  console.log('Duplicates found.')
+} else {
+  console.log('No duplicates found.')
+}
```
CosminPerRam, line 30: Thanks for this little tool, please add a newline at the end tho (:

CosminPerRam:
> Do you mind explaining me how this works? Thank you!

@Douile made GIDs tests for rust-gamedig, the code is [here](https://github.com/gamedig/rust-gamedig/blob/main/crates/id-tests/src/lib.rs).
While at it he also added support for testing node-gamedig gids (as we have talked that we would eventually also want to change gids in it).

> It fails on game names that has special chars:

Well, testing gids is quite hard to do, and apparently not all cases are well-processed (those few left ones that are failing), that'll require some adjustments/fixes for the code.
Fixing CI and (if needed) the last few ids could come in a separate PR.

lib/game-resolver.js
podrivo: Sorry, but you mean moving the `Object.keys(games).forEach()` iteration to `bin/gamedig.js`?

Douile:
The ID test are based off the rules in [CONTRIBUTING.md](https://github.com/gamedig/node-gamedig/blob/master/CONTRIBUTING.md#naming). They are only based of the ID and the name of each game so here it looks like we have the unfortunate situation of triggering a failure based on changes prior to this PR.

There are a few bugs shown here, I'll make an issue about them. The only actual bad ID looks like terraria which has a typo.

I don't think this test should block this PR though as the ID issues are not caused by it.

tools/find_id_duplicates.js
podrivo: Added! (:

lib/game-resolver.js
CosminPerRam: Maybe I haven't explained it that well
```js
if (!options.skipOldIDs) {
  Object.keys(games).forEach((id) => {
    if (games[id]?.extra.old_id) {
      game = games[id]
    }
  })
}
```

CONTRIBUTING.md
```diff
@@ -24,19 +24,23 @@ so.
...
-release year is already part of the name).  
+if not listed there) with also the release year included as `games.release_year`. 
 If there is a mod that needs to be added (or it adds the support for server
 queries for the game), its name should be composed of the game name, a separating
 **bracket**, the mod name and the release year as specified previously
-(e.g. `Grand Theft Auto V - FiveM (2013)`).
+(e.g. `Grand Theft Auto V - FiveM`).
+
+Older IDs exist because of a naming system update. Having them as an option, 
+guarantees that other softwares will still function after a library update.
```
CosminPerRam, line 39: This isn't really related to what this document is, the line could be added in the migrating document tho.

tools/generate_games_list.js
podrivo: Oh yeah, we've discussed this. Let me create this document as well.

If the idea is to eventually remove them, then yeah, let's not promote them. But what's the downside of having them as ids that used to work?

CosminPerRam: Yes, that's not a downside, i just meant to do not include them in this file.

lib/game-resolver.js
podrivo: Oh, yes, of course! haha Thank you! 😅

podrivo:
`MIGRATION.md` document was created. Please, let me know what else should I include there! 🙌

tools/generate_games_list.js
```diff
@@ -22,11 +22,12 @@ sortedGamesIds.forEach(key => {
 
 let generated = ''
 generated += '| GameDig Type ID | Name | See Also\n'
-generated += '|---|---|---\n'
```
CosminPerRam, line 25: Could you revert this file just as it was?

CosminPerRam: (also the games_list file because of this)

podrivo: Yeah, of course!

podrivo:
After a battle to roll back the file, I've managed to do it. Sorry about the mess...

CosminPerRam:
Thanks for taking care of this, looks good to merge.
> skipOldIDs (defaults to false).

I would change this to `true` by default as the old ids aren't present in the GAMES_LIST file and would create a bit of confusing (and again, we wouldn't want users to use the old ones), this (the mention to use that option when handling old ids) can be mentioned in the first lines of the migration doc. Thoughts?

podrivo:
> I would change this to true by default [...]

If we do this, it'll break implementations in Homepage or Uptime Kuma?

CosminPerRam:
Yes, but as seen in [Homepage#2474](https://github.com/gethomepage/homepage/pull/2474), [Homepage#1939](https://github.com/gethomepage/homepage/pull/1939), [Homepage#1886](https://github.com/gethomepage/homepage/pull/1886) where I updated them, (but in Uptime Kuma was done by someone else), the person that updates any dependency is expected to test and mention any breaking changes that could occur by updating.
In these cases we would mention this and disable the skipping of old ids because of this, for the sake of avoiding creating any breaking changes for them.

podrivo:
Hmm, I thought you said that using new ids as default was not possible: https://github.com/gamedig/node-gamedig/issues/493#issuecomment-1902171852
I'm not against changing, just a bit confused. 😅 

—

If skip checking old ids is going to be true by default, maybe we should rename it to `checkOldIDs`? 
So it's false by default.

```js
// Don't check by default
gamedig --type counterstrike2 45.144.155.107:27015

// Check old ids
gamedig --type counterstrike2 45.144.155.107:27015 --checkOldIDs
```

CosminPerRam:
Well I can't say that I'm explaining stuff very well in general, sorry for the confusion.
> maybe we should rename it to checkOldIDs

LGTM.

Regarding that comment:
> [...] so we should resolve it as smoothly as possible for them.

In this case, we should inform them that they need that option.

podrivo:
Should we also mention the migration document in our README?

CosminPerRam:
Yes

PR #497, opened 2024-01-21
Fix: Palworld results - Name of the server is empty

jammsen:
While ASSA returns something in CUSTOMSERVERNAME_s, Palworld returns that on epic as empty, but NAME_s is set with the servername.

Can be tested with:
npx gamedig --debug --type palworld 148.251.89.181:8211
npx gamedig --debug --type asa 148.251.89.181:7783
Both servers are only and from me.

@CosminPerRam Please feel free to improove quality on this fix, you know my skill-level on Javascript 😎

### DIFF
--- a/protocols/epic.js
+++ b/protocols/epic.js
@@ -126,18 +126,13 @@ export default class Epic extends Core {
       throw new Error('Server not found')
     }
 
-    if (
-      desiredServer.attributes.NAME_s !== undefined &&
-      desiredServer.attributes.NAME_s !== null &&
-      desiredServer.attributes.NAME_s !== ''
-    ) {
-      state.name = desiredServer.attributes.NAME_s;
-    } else if (
-      desiredServer.attributes.CUSTOMSERVERNAME_s !== undefined &&
-      desiredServer.attributes.CUSTOMSERVERNAME_s !== null &&
-      desiredServer.attributes.CUSTOMSERVERNAME_s !== ''
-    ) {
-      state.name = desiredServer.attributes.CUSTOMSERVERNAME_s;
+    const name = desiredServer.attributes.NAME_s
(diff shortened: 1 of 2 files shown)


protocols/epic.js
```diff
@@ -126,7 +126,21 @@ export default class Epic extends Core {
...
+      state.name = desiredServer.attributes.NAME_s;
+    } else if (
+      desiredServer.attributes.CUSTOMSERVERNAME_s !== undefined &&
+      desiredServer.attributes.CUSTOMSERVERNAME_s !== null &&
+      desiredServer.attributes.CUSTOMSERVERNAME_s !== ''
+    ) {
+      state.name = desiredServer.attributes.CUSTOMSERVERNAME_s;
+    } else {
+      state.name = '';
+    }
```
podrivo, line 143: Maybe we can have something like this?

```js
const name = desiredServer.attributes.NAME_s
const customName = desiredServer.attributes.CUSTOMSERVERNAME_s

if (name && name !== '') {
  state.name = name
} else if (customName && customName !== '') {
  state.name = customName
} else {
  state.name = ''
}
```

jammsen: Sure, fine with me. 
My JS is very old knowledge and therefore rusty, is just test and give an idea, good execution and looking beatiful, that i will leave up to you guys 😎

jammsen: Updated it

CosminPerRam:
Seeing #500, it would be better to move this name assigning to the protocol file (`palworld.js`).

It would look something like this:
```js
// this is a new function in that class
async run (state) {
    await super.run(state)
    state.name = state.raw.attributes.CUSTOMSERVERNAME_s;
}
```

CosminPerRam (approved):
Great little fix! Thanks.

jammsen:
My first fix 😆

PR #519, opened 2024-01-27
feat: add support to CommonJS

### LINKED ISSUE #517: feat: Add support for ESModules and CommonJS at the same time
**What is this feature about?**
Since gamedig is a library, it would be more interesting to allow users who use it to choose within the scope of their projects whether to use ESModules or CommonJS.

**Additional context/references**
Many people are having problems using version 5.0 because their projects use CommonJS `require("gamedig")`. Using a bundler like esbuild, it is possible to maintain the same code base and support both module methods.

I did some testing with esbuild, where the library is transpiled to ESModule and CommonJS with a script.

- Repo: https://github.com/TribufuForks/node-gamedig/tree/esbuild
- Diff: https://github.com/gamedig/node-gamedig/compare/master...TribufuForks:node-gamedig:esbuild

Just keep in mind that this adds an extra step to the lib, requiring the build script to be run before publishing, for example. But at the same time it removes the obligation to use ESModules.

GuilhermeWerner:
### This PR adds

Support for CommonJS modules using esbuild.

Closes #517

### Explanation

To support projects that use ESModules and CommonJS at the same time, it is possible to use a bundler to transpile the code.

In this PR, esbuild was used, available in the file `tools/esbuild.js`, when executing this file a lib bundle converted to CommonJS is generated, which can then be used by projects that use `require`.

So that the npm package knows which file should be used according to the project type, the `main` directive of `package.json` was replaced by `exports`, which determines the two files that will be exported according to the type.

```json
"exports": {
    "import": "./lib/index.js",
    "require": "./dist/index.cjs"
},
```

The bundler settings can be changed in the script, in this PR the transpiled file is created in the `dist/index.cjs` folder, however it is up to the lib maintainers to decide where this file should be, just remember that this file was generated by the bundler and It has relatively low readability and should be ignored from the repository.

To facilitate the package publishing process on NPM, scripts were created to run the build automatically when the package is installed (prepare).

The examples were also changed to demonstrate use with require and import.

### Detail about the got dependency

The http client library that is used only supports ESModules (https://github.com/sindresorhus/got/issues/1789).
image

To fix this, it is included in the bundle so that esbuild can convert its source code. The downside of this is that the bundle file becomes much larger.


examples/require.cjs
```diff
@@ -0,0 +1,13 @@
+const { GameDig } = require('../dist/index.cjs')
+// Instead of '../lib/index.js' you would have here 'gamedig'.
```
CosminPerRam, line 2: `../lib/index.js` should be `../dist/index.cjs` here.

package.json
```diff
@@ -3,7 +3,9 @@
     "description": "Query for the status of any game server in Node.JS",
     "scripts": {
         "lint:check": "eslint .",
-        "lint:fix": "eslint --fix ."
+        "lint:fix": "eslint --fix .",
+        "prepare ": "npm run build",
```
CosminPerRam, line 7: There is a trailing whitespace after the word `prepare`.

tools/esbuild.js
```diff
@@ -0,0 +1,15 @@
+import { build } from 'esbuild'
+import { nodeExternalsPlugin } from 'esbuild-node-externals'
+
+const buildConfig = {
+  entryPoints: ['lib/index.js'],
+  platform: 'node',
+  target: 'node14',
```
CosminPerRam, line 7: target should be node16 as we have moved to it.

tools/esbuild.js
```diff
@@ -0,0 +1,15 @@
...
+  bundle: true,
+  outfile: 'dist/index.cjs',
+  format: 'cjs',
+  // got is a esm module, so we need to add it to the allowList, to include it in the bundle.
+  plugins: [nodeExternalsPlugin({ allowList: ['got'] })]
+}
+
+build(buildConfig).then(() => { }).catch(() => { process.exit(1) })
```
CosminPerRam, line 15: Please add a trailing newline (:

CosminPerRam (changes requested):
LGTM, thanks bunch.

Edit: forgot to mention to add a line in the changelog file.

CosminPerRam:
Thanks bunch, just a note to have them for future references:
> Detail about the got dependency

This is unfortunate, but we'll have to go with this for now.

I too think that the abrupt move to ESM wasn't exactly the best choice for v5, but I think it'll be fair to completely remove support for it for v6 (as by then we would already be passing 3 years since stable esm support).

CosminPerRam (changes requested):
Forgot to mention to add a line in the changelog file

GuilhermeWerner:
> Forgot to mention to add a line in the changelog file

I always forget the changelog 🫠

podrivo:
@GuilhermeWerner Sorry, a last minute conflict... :(

CosminPerRam (approved):

Issue #520, opened 2024-01-28
bug: Palworld player array empty


DaliborTrampota:
**Describe the bug**
response.players is empty array even though players are joined

**Screenshots or Data**
![image](https://github.com/gamedig/node-gamedig/assets/31895316/c711c9f2-915f-4ad4-b9b7-534e2684d0e7)

**Additional context**
Im on master due to the beta.0 not showing numplayers 0 even when there are players joined
"gamedig": "github:gamedig/node-gamedig"

podrivo:
Hello @DaliborTrampota. This has been fixed, but not yet published.
Currently, you can grab players from `state.raw.attributes.PLAYERS_`

Related https://github.com/gamedig/node-gamedig/issues/507 and https://github.com/gamedig/node-gamedig/pull/508

Maxence1502:
I think he wants to get the player names instead of the number

podrivo:
Palworld's protocol (EOS) does not provide players data.

PR #522, opened 2024-01-29
chore: Add EOS notes on Palworld and The Isle Evrima

podrivo:
This adds EOS Notes on Palworld and The Isle Evrima.
Closes this issue: https://github.com/gamedig/node-gamedig/issues/520

—

Also, would like to have more input on a better solution for detecting this protocol on `tools/generate_games_list.js`
It's currently like this:

```
if (
  game.options.protocol === 'epic' || 
  game.options.protocol === 'asa' || 
  game.options.protocol === 'palworld' || 
  game.options.protocol === 'theisleevrima'
) {
  notes.push('[EOS Protocol](#epic)')
}
```

### DIFF
--- a/GAMES_LIST.md
+++ b/GAMES_LIST.md
@@ -205,7 +205,7 @@
 | openarena            | OpenArena                                        |                                                 |
 | openttd              | OpenTTD                                          |                                                 |
 | painkiller           | Painkiller                                       |                                                 |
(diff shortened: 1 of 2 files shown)


CosminPerRam (approved):
Thanks bunch, that `if` is getting quite long, we'll have to see about that, but its fine for now.

PR #526, opened 2024-01-31
Updated the eco protocol to include player names and fix problems encountered when using the eco web interface behind a proxy.

Vito0912:
This is my first pull request for this project. So if something is not aligning with the guidelines or coding styles, please give me a hint so I can change it.

The existing code for the Eco game functions correctly. However, it does not use any server protocol; instead, it accesses the web interface (just a plain http request). Many servers employ reverse proxies (or other methods) that cannot connect via the ip. A straightforward solution is to use the host name rather than the IP address. I'm uncertain why the IP address was originally used, given that this is merely a web request.

Additionally, I have included player information in the state.

### DIFF
--- a/protocols/eco.js
+++ b/protocols/eco.js
@@ -5,16 +5,17 @@ export default class eco extends Core {
     if (!this.options.port) this.options.port = 3001
 
     const request = await this.request({
-      url: `http://${this.options.address}:${this.options.port}/frontpage`,
+      url: `http://${this.options.host}:${this.options.port}/frontpage`,
       responseType: 'json'
     })
     const serverInfo = request.Info
 
     state.name = serverInfo.Description
-    state.numplayers = serverInfo.OnlinePlayers;
+    state.numplayers = serverInfo.OnlinePlayers
     state.maxplayers = serverInfo.TotalPlayers
     state.password = serverInfo.HasPassword
     state.gamePort = serverInfo.GamePort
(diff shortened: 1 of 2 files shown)


CosminPerRam (changes requested):
LGTM! Thanks for the hotfix, please also make sure to add a line in the changelog file (:

protocols/eco.js
```diff
@@ -5,16 +5,17 @@ export default class eco extends Core {
...
     })
     const serverInfo = request.Info
 
     state.name = serverInfo.Description
-    state.numplayers = serverInfo.OnlinePlayers;
+    state.numplayers = serverInfo.OnlinePlayers
     state.maxplayers = serverInfo.TotalPlayers
     state.password = serverInfo.HasPassword
     state.gamePort = serverInfo.GamePort
+    state.players = serverInfo.OnlinePlayersNames ? serverInfo.OnlinePlayersNames.map(name => ({ name, raw: {} })) : []
```
CosminPerRam, line 18: You can do `serverInfo.OnlinePlayersNames?.map(name => ({ name, raw: {} })) || []`, which would be shorter (:

Vito0912: Pushed a new version

Vito0912:
Used `serverInfo.OnlinePlayersNames?.map(name => ({ name, raw: {} })) || []` for readability and added line in changelog file

CosminPerRam (approved):

Issue #543, opened 2024-02-15
bug: TeamSpeak2 server name not showing


podrivo:
**Describe the bug**
Querying a TeamSpeak2 have no name, despite having the name in the `state.raw.server_name`.

**Screenshots or Data**
Query response:

```
gamedig --type protocol-teamspeak2 130.44.23.184:9112 --pretty

{
  "name": "",
  "map": "",
  "password": false,
  "raw": {
    "server_id": "1",
    "server_name": "*\\0/* #PALMWAYGIRLS *\\0/*",
    "server_platform": "Win32",
    "server_welcomemessage": "The TS3 will remain open for Misc. Live Shows. Voice Cheer Chat will still be Avalible for those that can't be in Studio. TS2 is our Backup Voice Chat to Studio...",
    "server_webpost_linkurl": "",
    "server_webpost_posturl": "",
    "server_password": "1",
    "server_clan_server": "0",
    "server_udpport": "9112",
    "server_maxusers": "5",
    "server_allow_codec_celp51": "0",
    "server_allow_codec_celp63": "0",
    "server_allow_codec_gsm148": "0",
    "server_allow_codec_gsm164": "0",
    "server_allow_codec_windowscelp52": "0",
    "server_allow_codec_speex2150": "0",
    "server_allow_codec_speex3950": "0",
    "server_allow_codec_speex5950": "0",
    "server_allow_codec_speex8000": "0",
    "server_allow_codec_speex11000": "0",
    "server_allow_codec_speex15000": "0",
    "server_allow_codec_speex18200": "0",
    "server_allow_codec_speex24600": "-1",
    "server_packetssend": "42008031",
    "server_bytessend": "5095618654",
    "server_packetsreceived": "35229703",
    "server_bytesreceived": "2732893069",
    "server_uptime": "32403181",
    "server_currentusers": "1",
    "server_currentchannels": "4",
    "server_bwinlastsec": "0",
    "server_bwoutlastsec": "23",
    "server_bwinlastmin": "19",
    "server_bwoutlastmin": "24",
    "channels": [
      {
        "id": "1",
        "codec": "12",
        "parent": "-1",
        "order": "0",
        "maxusers": "5",
        "name": "CATS HANGOUT",
        "flags": "0",
        "password": "0",
        "topic": "Classmates and Cheer Lobby"
      },
      {
        "id": "2",
        "codec": "12",
        "parent": "-1",
        "order": "20",
        "maxusers": "10",
[...]

PR #544, opened 2024-02-15
fix: TeamSpeak2 server name

podrivo:
Fix how TeamSpeak 2 protocol set the `state.name`.
It closes this issue https://github.com/gamedig/node-gamedig/issues/543


CosminPerRam (approved):
Thanks!

Issue #388, opened 2023-10-23
Support for open.mp


ReshiramZekrom1:
**What is this feature about?**
Adding [open.mp](https://github.com/openmultiplayer/open.mp) support to gamedig

**Additional context/references**
open.mp is currently an improved version of SA-MP, made entirely from scratch and open-source. Even their web page with the server fetcher is open-source. I'll link the repository [here](https://github.com/openmultiplayer/web/tree/master/app) because maybe it could be of some use.

Thanks

CosminPerRam:
Some research will be required to add support for the game (meaning low priority), the project seems very interesting.

ReshiramZekrom1:
> Some research will be required to add support for the game (meaning low priority), the project seems very interesting.

Thanks for the reply! 

I've found out they use the same [query mechanism](https://sampwiki.blast.hk/wiki/Query_Mechanism#Response_from_c) that SA-MP uses except for the "d" packet which isn't implemented (SA:MP servers got DoS attacked with it apparently). Anyways the "c" packet is working although it isn't possible to get the player ID and the score with it.

CosminPerRam:
Hey, it seems that they do respond to that query (queried random servers from the server list on their website), do you have any references of this not working (maybe where has been discussed about it?).
I linked the support PR to this issue, I will merge it when this matter gets resolved.

ReshiramZekrom1:
The server doesn't respond to the "[d query](https://github.com/gamedig/node-gamedig/blob/17c351fb4bc5e32d5426e6ca9cc55c67589531ff/protocols/samp.js#L54)" because that request is not supported. With gamedig the server results online but you cant get the current players list. If you replace 'd' with the 'c' in that line

CosminPerRam:
Hmm, but its weird because the server are responding to it, so I guess its normal behaviour?
Query with the 'd' query enabled on a server from their master server:
![image](https://github.com/gamedig/node-gamedig/assets/7972857/4c2bbfad-9ed5-4ad8-8d00-da44bd7627e1)
Without that query:
![image](https://github.com/gamedig/node-gamedig/assets/7972857/21795265-5be0-4f2a-bdfc-f9a459762cad)

Regarding the second screenshot, gamedig does not query with the c/d packet if on the server are more than 100 players (as the server wont respond for possible the reason that you stated).

ReshiramZekrom1:
[...]

PR #547, opened 2024-02-22
Added c query requests for gtasao (open.mp)

Focus04:
This is a follow-up to #388. User who opened the issue was right and has been backed up by various open.mp devs. `D` query is not yet supported by omp, which breaks regular `gtasam` queries on these servers, resulting in an empty `players` array. A workaround is using `C` queries instead, similarly to how `vcm` is handled, which returns a less detailed, but handy array of players along with their scores.
This adds a separate protocol for the already existing `gtasao` query type, which extends base `samp`. This new protocol sends a `C` packet instead of `D` and will consequently return a populated array of players, as expected.
Below are 2 queries for a server running open.mp with these changes added. More of a _before and afer_ since both queries previously returned empty players arrays.

- ### gtasam (D query - empty player list)

image (2)

- ### gtasao (C query - populated player list)
image (1)


protocols/gtasao.js
```diff
@@ -0,0 +1,8 @@
+import samp from './samp.js'
+
+export default class gtasao extends samp {
+  constructor() {
+    super()
+    this.isOmp = true
+  }
+}
```
CosminPerRam, line 8: Please add a trailing newline here (:

CosminPerRam (changes requested):
LGTM, please also add a line in the changelog file.

Issue #536, opened 2024-02-05
feat: Add status object player's online time


Mululu:
**What is this feature about?**
A status object with a player's online time would be desirable. The RAW output does include the time, but this designation varies depending on the game (time, uptime, ...) so this data can only be queried to a limited extent.

**Additional context/references**
Example:

RAW output: Conan Exiles = time
RAW output: Farming Simulator 22 = uptime

CosminPerRam:
In order to stabilize a field (meaning to add one), the majority of the protocols would have to provide it and unfortunately I don't think many do that (cannot check know, will do later).

podrivo:
@Mululu Hello! Can you give us examples of games that provide the data you mentioned? Thank you!

CosminPerRam:
@podrivo he mentioned under "examples" Conan Exiles, which do provides `time` for players.
FS22 could also but I can't remember its raw output to know.

CosminPerRam:
> But that said, shouldn't the valve protocol contain the uptime

It does, but many games (that uses the valve protocol) do not provide that data.

Mululu:
Can this work?

`   $('Player').each(function () {
  if ($(this).attr('isUsed') === 'true') {
    state.players.push({
      name: $(this).text(),
      **time: parseInt($(this).attr('uptime'), 10),**
      raw: {
        isAdmin: $(this).attr('isAdmin') === 'true'
      }
    })
  }
})`

Vito0912:
Ofc this can work (Just formatted below)
```
    $('Player').each(function () {
      if ($(this).attr('isUsed') === 'true') {
        state.players.push({
          name: $(this).text(),
          raw: {
            isAdmin: $(this).attr('isAdmin') === 'true',
            uptime: parseInt($(this).attr('uptime'), 10)
          }
        })
      }
    })
```
To
```
    $('Player').each(function () {
      if ($(this).attr('isUsed') === 'true') {
        state.players.push({
          name: $(this).text(),
          time: parseInt($(this).attr('uptime'), 10)
          raw: {
            isAdmin: $(this).attr('isAdmin') === 'true'
            uptime: parseInt($(this).attr('uptime'), 10)
          }
        })
      }
    })
```

But it is not within the scope. As @CosminPerRam outlined,

> the majority of the protocols would have to provide it.

It must (or should) be supported by many games, and that's currently not the case, as far as I know. If you were to stabilize fields, there should be a certain number of games that feature a field like this.
[...]

PR #549, opened 2024-02-24
feat: FS add player raw time

CosminPerRam:
Adds a 'time' field to the raw values of a player in the farming simulator protocol.
This has been sourced from #536.


protocols/farmingsimulator.js
Vito0912, line 1: Wouldn't it be more good of a standard if we only use time instead of both?
It's a bit redundant.

CosminPerRam:
Oh, I am blind, this indeed isn't needed.

PR #560, opened 2024-04-08
feat: add palworld new new query implementation

### LINKED ISSUE #555: Palworld broken after update

CosminPerRam:
This replaces #556, although I haven't tested it to see if that endpoint still works, could be added eventually as a master server query, but I'll close it for now.

Finally closes #555.

### DIFF
--- a/lib/games.js
+++ b/lib/games.js
@@ -2092,7 +2092,7 @@ export const games = {
     name: 'Palworld',
     release_year: 2024,
     options: {
-      port: 8211,
+      port: 8212,
       protocol: 'palworld'
     }
   },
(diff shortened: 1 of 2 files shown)


Destarianon:
How long does it take for a change like this to make it into the NPM package available for use?

CosminPerRam:
It'll become available on NPM when we'll make a new release.
But until then you can install the library from this repository (therefore having this available now), just do `npm i git+https://github.com/gamedig/node-gamedig` to install gamedig with the latest fixes.

PR #562, opened 2024-04-29
feat: add `Nova-Life: Amboise` support

CosminPerRam:
I couldn't find the official default query port so I'll let it be the default protocol one for now.


Maxence1502:
Default query port is 27015, that's correct

PR #563, opened 2024-05-07
Adding Abiotic Factor support

RStivanson:
Adds Abiotic Factor to the list of games.


CosminPerRam (approved):
Tried on a couple of servers and everything looks good, thanks!

Issue #571, opened 2024-06-06
feat: Support Minetest


lotigara:
###### please sorry for my English

**What is this feature about?**
I want to use Uptime Kuma (it is an uptime monitor that supports GameDig) with my own Minetest server, but Uptime Kuma doesn't support UDP (it is what Minetest uses), and GameDig doesn't support Minetest.

**Additional context/references**
Minetest uses UDP port 30000, there is my server: `lotigara.ru`

PR #573, opened 2024-06-09
feat: Add Minetest support

xCausxn:
Added support for minetest, which does just use the serverlist under the hood.

Implements initial version for #571


lotigara:
I hope that this PR will be merged!

CosminPerRam (approved):
Looks great, tried to query a few servers and everything works as expected, thanks for the contribution!

Issue #49, opened 2016-12-20
Support for RAGE MP


kotbrain:
Multiplayer for GTA 5 will be released soon. It will also be possible to get information about the servers there. Will you implement multiplayer support and how soon?

https://www.youtube.com/watch?v=gs_2zVSDLzw

sonicsnes:
Is this different from FiveM?

twamp22:
Appears to be a different mod. RageMP seems to be a JS/NodeJS client/server, whereas FiveM is mainly C variants. RageMP is however closed source it would seem, while FiveM is open source.

sonicsnes:
This game does not seem to have a query protocol at all unfortunately 😢. Someone would need to write a plugin for rage.mp which keeps track of the connected players and exposes it in some way.

eduardozgz:
RageMP seems to use these endpoints for querying servers on their official website (https://rage.mp/servers and https://rage.mp/masterlist/):
- https://cdn.rage.mp/master/
- https://cdn.rage.mp/master/v2/ (similar to the previous one, but with the ability to group servers)

I have been using the first endpoint for a few years, and it works fine. If you are okay with querying the servers using these endpoints, I would be happy to make a PR. I will implement this in a fork for my Discord bot, as I need it for my own use.

CosminPerRam:
As we already have a bunch of games that rely on master servers to get a server's informations, this could be done for ragemp too, opening for further research.

xCausxn:
V2 Endpoint is grouped so in order to find a specific server is could either be:
`key === ip:port` or needing to loop through each `servers` array in each key to find it.

**Example**
```json
{
    "id": "grandrpen",
    "servers": [
      {
        "id": "rage2.gta5grand.com:22005",
        "name": "🛑[EU][US][voice] [roleplay] Grand RolePlay | Server 2 | discord.gg/gta5grandcom [1.1]",
        "lang": [
          "en"
        ],
        "players": {
          "amount": 789,
          "max": 3000,
          "peak": 797
        }
      },
      {
        "id": "rage3.gta5grand.com:22005",
        "name": "🛑[EU][US][voice] [roleplay] Grand RolePlay | Server 3 | discord.gg/gta5grandcom [1.1]",
        "lang": [
          "en"
        ],
        "players": {
          "amount": 775,
          "max": 5000,
          "peak": 775
        }
      }]
}
```

However the id could also be domain:port rather than ip:port

[...]

PR #576, opened 2024-06-19
feat: Add Grand Theft Auto V - RAGE MP support

xCausxn:
Implements support for getting server info from the rage mp server list.

Completes: #49


xCausxn:
@CosminPerRam I opted to just return the entry server result entry as the "raw" field. Partly due to the fact potential info for the nested entry is available at the top level.

We could however remove excess entries and move the servers[X] entry to the top level for our case.

CosminPerRam (approved):
Great stuff!

CosminPerRam:
> Partly due to the fact potential info for the nested entry is available at the top level.

We can leave it this way.

CHANGELOG.md
```diff
@@ -392,4 +393,4 @@ during the query.
 
 ### 1.0.0
 * First official release
-* Node.js 6 is now required
+* Node.js 6 is now required
```
CosminPerRam, line 396: Hmmm, seems like a stray change here?

lib/games.js
```diff
@@ -1270,6 +1270,14 @@ export const games = {
...
     }
   },
+  gta5r: {
+    name: 'Grand Theft Auto V - RageMP',
+    release_year: 2016,
+    options: {
+      port: 22005,
+      protocol: 'ragemp'
+    }
+  },
```
CosminPerRam, line 1280: Please add this to properly make the table generator reference the note:
```
extra: {
  doc_notes: 'gta5r'
}
```

Like as in:
```
  trackmaniaforever: {
    name: 'Trackmania Forever',
    release_year: 2008,
    options: {
      port: 2350,
      port_query: 5000,
      protocol: 'nadeo'
    },
    extra: {
      doc_notes: 'nadeo'
    }
  },
  ```
```

CosminPerRam (changes requested):

CosminPerRam (approved):

Issue #587, opened 2024-07-09
feat: Add Alt: V support


dgibbs64:
**What is this feature about?**
Found a GTA V Mulitplayer mod called Alt:V https://altv.mp/ looks like it has a master server API that might be able to query to get server details https://docs.altv.mp/articles/master_list_api.htm

Thought you guys might like to add it

xCausxn:
It does seem like another weird one, as the address field has an ip:port but could be different to the port field. Another thing is this could be a domain:port based field as well.

A similar approach to ragemp could be done, but it will have some assumptions made.

Another approach could be similar to how we use `options.guildId` for discord server queries?

PR #588, opened 2024-07-09
Add support for AltvMP

xCausxn:
This PR adds support for AltvMP. (#587)

It implements two methods to find the server using the master list at https://api.alt-mp.com/servers

If the `serverId` argument is passed, then a direct lookup can be done at /servers/ID

This does also add a fix to state.connect in general, as if the connect field is manually set, core.js just overrides it. (likely something that was missed when refactored)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 5.X.Y
 * FOUNDRY - Added support (#585)
 * Grand Theft Auto V: AltVMP - Added support (By @xCausxn #588)
+* Fix ability to manual set `state.connect` (By @xCausxn #588)
 
 ## 5.0.1
 * Minetest - Added support for minetest utilizing official server list (By @xCausxn #573)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## To Be Released...
 ## 5.X.Y
 * FOUNDRY - Added support (#585)
+* Grand Theft Auto V: AltVMP - Added support (By @xCausxn #588)
 
 ## 5.0.1
 * Minetest - Added support for minetest utilizing official server list (By @xCausxn #573)
--- a/protocols/core.js
+++ b/protocols/core.js
@@ -84,7 +84,7 @@ export default class Core extends EventEmitter {
     state.queryPort = options.port
     // because lots of servers prefix with spaces to try to appear first
     state.name = (state.name || '').trim()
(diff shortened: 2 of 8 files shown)


xCausxn:
@CosminPerRam a potential implementation that adds some flexibility. 

Seeing as we do `guildId` for discord, seemed like an ok addition. We could even just look to change `guildId` to also be `serverId` (with backwards compat of course) as its a bit more generic and could apply to multiple servers implementations in the future.

CosminPerRam:
Looks great, I really like the option to use the id or the host:port but could you argue the addition of the `useCdn` option? In [their docs](https://docs.altv.mp/articles/cdn_links.html), its just a CDN, so it *could* be used for multiple servers, right? I haven't checked with a script to see if there are multiple entries with the same cdn url, but could it be possible to do so?

protocols/core.js
```diff
@@ -84,7 +84,7 @@ export default class Core extends EventEmitter {
     state.queryPort = options.port
     // because lots of servers prefix with spaces to try to appear first
     state.name = (state.name || '').trim()
-    state.connect = `${state.gameHost || options.host || options.address}:${state.gamePort || options.port}`
+    state.connect = state.connect || `${state.gameHost || options.host || options.address}:${state.gamePort || options.port}`
```
CosminPerRam, line 87: This is technically a breaking change and a bug, currently `discord`, `eldewrito` and `buildandshoot` do (or may) set this field manually and although the state should use that one, it does not and manually sets one (via the fields before editing).

It is a breaking change as users of these protocols could get different data after updating to a newer version.
And it is a bug as the field should respect the protocol's selection and not overwrite this.

CosminPerRam:
> Seeing as we do guildId for discord, seemed like an ok addition. We could even just look to change guildId to also be serverId (with backwards compat of course) as its a bit more generic and could apply to multiple servers implementations in the future.

For sure, great idea, please make a tracking issue for this (:

protocols/core.js
CosminPerRam: Talked over on discord and this should be fine.

CosminPerRam (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## To Be Released...
 ## 5.X.Y
 * FOUNDRY - Added support (#585)
+* Grand Theft Auto V: AltVMP - Added support (By @xCausxn #588)
```
CosminPerRam, line 6: Actually, please mention the `state.connect` as a fix here.

PR #597, opened 2024-08-03
feat: add a rules bytes return for valve protocol

blackwaterbread:
Valve protocol seems to need an option to return rules as bytes.


CosminPerRam (approved):
Sure, we could have something like this.

PR #610, opened 2024-08-16
fix: registerRTT would break successful queries

CosminPerRam:
@mgrimace encountered issues by using the [Homepage](https://github.com/gethomepage/homepage/) project and adding a widget which uses `minecraft`, reported [here](https://github.com/gethomepage/homepage/discussions/3862).

After some debugging we have cloncluded that the bedrock query runs (from logs, as the target server is bedrock edition) and other fails, but the query still fails.

~~Commit [9bd3caa](https://github.com/gamedig/node-gamedig/commit/9bd3caab7b8b45d20349416419ddc36390325782) is the only a contester as why this happens, OP verified this by making the query with the version before this change (`5.0.0-beta.2`) which worked.~~ edit: it was not

Edit:
Commit [6b85c6b](https://github.com/gamedig/node-gamedig/commit/6b85c6bc0b1b3e41dcf3350499af4b21ec43da5e) ~~is also a possiblity~~ (late edit) caused successful queries to fail if `registerRtt` would take too long.

This PR pretty much reverses that commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## To Be Released...
 ## 5.X.Y
 * Added Vintage Story support via the master server (#606)
+* Fixed `registerRtt` breaking successful queries if it didn't respond in the query timeout (#610)
 * Added support for rFactor 2 (By @xCausxn #614)
 
 ## 4.3.2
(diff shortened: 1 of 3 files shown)


CosminPerRam:
A server where I have observed this behaviour: `play.pokebedrock.com:19132`

CosminPerRam:
@cetteup made a great explaination and summary of this in #611

PR #627, opened 2024-09-03
feat: remove cheerio dependency

CosminPerRam:
`cheerio` was added for `Build And Shoot` and was later used for `Farming Simulator`.
In the past, `Build And Shoot` had a request and regex implementation, and we are switching back to it (modified version, not original).
`Farming Simulator` seems to be a bit troublesome to be dealt with, remains to be seen.

According to this Bundlephobia scan of the package.json:
image
Shows that (without `got` at least), `cheerio` is almost half of the package size.
This wouldn't really be a trouble for modules, but it is for commonjs as its pulling the entire package.

### DIFF
--- a/protocols/buildandshoot.js
+++ b/protocols/buildandshoot.js
@@ -1,5 +1,4 @@
 import Core from './core.js'
-import * as cheerio from 'cheerio'
 
 export default class buildandshoot extends Core {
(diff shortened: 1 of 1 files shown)


CosminPerRam:
Replaced by #683

Issue #607, opened 2024-08-13
ci: check or alter GAMES_LIST.md


CosminPerRam:
By adding support for a game, the `GAMES_LIST.md` file needs to be updated accordingly.

Doing this manually is quite error prone (wrong alphabetical order or forgetting to update references) and this results in PRs that do run the `tools/generate_games_list.js` having unnecessary extra changes (#606 is the latest example as of writing where we added support for Vintage Story).
![image](https://github.com/user-attachments/assets/5a82b390-d953-4630-ac92-3c2db9635622)
![image](https://github.com/user-attachments/assets/da993084-1e65-4d7f-ae3f-8719408d0a42)

This can be easily fixed by having CI that checks if the document is rightfully formatted or to forcefully update it automatically.

The `tools/generate_games_list.js` script has to be altered a bit, as running it right now adds the new table at the top, not eliminating the old one.

CosminPerRam:
Oops, didnt mean to close.

PR #635, opened 2024-09-15
fix: Generate games list with padded delimiter

RattleSN4K3:
Adresses the issue #607. Script will generate the same output as the current list of games.

Anchors are added back to `GAMES_IST.md`, to mark the start and end of the list (to be replaced).

### DIFF
--- a/GAMES_LIST.md
+++ b/GAMES_LIST.md
@@ -1,4 +1,6 @@
 ### Supported
+
+<!--- BEGIN GENERATED GAMES -->
 | GameDig Type ID      | Name                                             | See Also                                        |
 |----------------------|--------------------------------------------------|-------------------------------------------------|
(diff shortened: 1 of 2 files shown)


CosminPerRam (commented):
Works as expected, thanks bunch!
Technically this PR does not close #607 (at least completely) as a github workflow is needed to check (and/or overwrite) it.
Would you like to have a try at doing this also? I could merge this and do the CI part separately, but just asking if you're interested (:

RattleSN4K3:
Yeah. I noticed that a CI part is mentioned as well. But I don't think I'm able to fully create the action with every step it requires (like pushing) - less experience with it. Other than [copying an existing action](https://gist.github.com/RattleSN4K3/491b81ebcb60ec8f1dbdb92265b43677) and adjusting some parts, I would forward that to you, @CosminPerRam. 😀

Issue #631, opened 2024-09-10
feat: Adding UT3 LAN hosted Server as Supported Game


Hornochs:
**What is this feature about?**
Adding Unreal Tournament 3 which are hosted as LAN Match only as supported games, since LAN hosted servers doesn't use the GameSpy protocol for querying

**Additional context/references**
As in work in #628 a new protocol is needed. For structure reason all findings are written in #628

Issue #629, opened 2024-09-05
feat: Add Renegade X as supported Game


Hornochs:
**What is this feature about?**
Adding Renegade X for supported Games

**Additional context/references**
Uses the Unreal 3 / UDK Engine. All findings in regards the protocol are in #628

Issue #628, opened 2024-09-05
feat: Adding Toxikk as supported Game


Hornochs:
**What is this feature about?**
Add Toxikk, an Areana FPS as supported Protocol. 

**Additional context/references**
Toxikk uses Unreal 3 / UDK Engine with additional Support for the Valve Protocol.

Hornochs:
UDK Games using different SubSystems, for the networking stuff as well. In the Engine itself are systems integrated for LAN Servers and for Steam. The Developer can decide if Subsystems like steam are enabled or not. Toxikk uses Steam.

The Caviot: The Steam Subsystem gets only loaded on Dedicated Servers which are ment for Internethosting, equals startparamenter `bIsLanMatch=false` and make it possible to query by Valve Protocol. As soon as the servers are startet with `bIsLanMatch=true` the Steam Subsystem won't be loaded. Then a different Subsystem gets loaded, if I unterstand correctly it's `[IpDrv.OnlineGameInterfaceImpl]`

Having a look in Toxikk at `UDKGame/Config/UDKEngine.ini` seeing two references on Port 14001:

```
[IpDrv.OnlineGameInterfaceImpl]
LanAnnouncePort=14001
LanQueryTimeout=5.0
LanPacketPlatformMask=1
LanGameUniqueId=0x4D5707DB

[OnlineSubsystemLive.OnlineSubsystemLive]
LanAnnouncePort=14001
VoiceNotificationDelta=0.2
NumLogins=1
CurrentNotificationPosition=NNP_BottomRight
MaxLocalTalkers=2
MaxRemoteTalkers=9
bShouldLogArbitrationData=true
bShouldLogStatsData=true
LanQueryTimeout=3.0
LanPacketPlatformMask=1
LanGameUniqueId=1297287213
bShouldUseMcp=false
```

After digging in the Source Code of the Unreal Engine `[OnlineSubsystemLive.OnlineSubsystemLive]` referes to xBox 360 System. More Details about LAN Quering is written in [the officiel UDK documentation](https://docs.unrealengine.com/udk/Three/OnlineSubsystemSteamworks.html#LAN%20advertising/querying). After messing around I noticed the following "flow" how UDK handles LAN Discovery.

- Client sends an Broadcast Package from Source and Destination Port UDP 14001
[...]

Issue #608, opened 2024-08-13
feat: add alternative query for Vintage Story


CosminPerRam:
**What is this feature about?**
Altough we added support for this game in #606, it has been done via using the master server method, and [there is a mod](https://mods.vintagestory.at/theserverstats) for it to expose a query endpoint.

**Additional context/references**
It wouldn't hurt to provide more query methods, as master servers will eventually be shutdown.
The mod exposes a HTTP server and provides a GET endpoint which returns a json with the data, it seems pretty straight forward to add.

Original issue request for the game: #328

dgibbs64:
Small update. Turns out Killing Floor has 3 different querys. It also has gamespy1

tyronx:
> It wouldn't hurt to provide more query methods, as (for Vintage Story) master servers will eventually be shutdown.

Yep, this feature of yours incurred cloud hosting costs of over 5000 USD because you query this stuff so often. 96% of all queries on the master server is by gamedig. Jesus christ people. We will use every possible method available to us to block this service.

podrivo:
OMG @tyronx! Tell us more about this! :O

PR #638, opened 2024-09-15
[WIP] feat: UE3 LAN query support

RattleSN4K3:
This will add support for querying LAN servers for UE3 based games as issued by #631. But before merging this pull request, I like to support respectively work on another game, like _Toxikk_ and/or _Renegade X_.

This request has also various changes which might be be revised/checked. This pull request should be compatbile to all other protocols, but as the core is changed, it's a slightly critical change.

Additional related issues: #629 #628 

---

By default the LAN query will provide the same structure as a (master/online) server does. However as the _query provider_ is different with a LAN query the data is limited and will not provide any player data.

There is an additional _settings resolver_ in `patterns/ue3pattern.js` which currently holds the structure defined by UT3 and allows parsing the data in a more _structured_ way and also providing being able to customize any given payload/pattern by a simple JSON passed as parameter (WIP).

A setup for a query looks like this (for now) + some options for configuring the output:
```js
const gamedig = new GameDig({
  listenUdpPort: 14001
})

gamedig.query({
  type: 'unrealtournament3lan',
  address: '192.168.56.56'

  //port: 14002
  // outputAsArray: true, // instead of a single JSON item, this will output an array

  // packetVersion: 5,
  // lanGameUniqueId: '0x4D5707DB',
  // lanPacketPlatformMask: 1,

  // NOT IMPLEMENTED ANYMORE 
  // subsystem: 'lan'
  // usePattern: true // whether to use the new pattern logic by resolving the settings from a response into known properties
  // includeFlat: false, // [Pattern] if the root object should have all properties included
  // includeAll: false, // [Pattern] if the root object should have all properties included (in combiniation with includeLegacy=false and the property might be removed)
  // includeLegacy: true, // [Pattern] if the root object should include legacy/gamedig properties
  // includeNull: false, // [Pattern] whether the property should be included even if a mapped value cannot be resolved
})
```

An example response with the options `usePattern: true` and `outputAsArray: true`:
> [!WARNING]
> Outdated json response. See here for an updated https://gist.github.com/RattleSN4K3/55958bfc0639c4ce2347f5298cd70c6a
```js
[
  {
    "name": "Player",
    "version": "",
    "maxplayers": 8,
    "numplayers": 0,
    "players": [],
    "bots": [],
    "queryPort": 14001,
    "NumOpenPublicConnections": 8,
    "NumOpenPrivateConnections": 0,
    "NumPublicConnections": 8,
    "NumPrivateConnections": 0,
    "bShouldAdvertise": true,
    "bIsLanMatch": true,
    "bUsesStats": true,
    "bAllowJoinInProgress": true,
    "bAllowInvites": true,
    "bUsesPresence": true,
    "bAllowJoinViaPresence": true,
    "bUsesArbitration": false,
    "bAntiCheatProtected": true,
    "OwningPlayerName": "Player",
    "OwningPlayerId": "0000000000000000",
    "[server]": {
      "BotSkill": "Experienced",
      "PureServer": "Yes",
      "LockedServer": "No",
      "AllowKeyboard": "No",
      "IsFullServer": "No",
      "IsEmptyServer": 1,
      "IsDedicated": "Yes",
      "SteamID": 0,
      "ServerDescription": ""
    },
    "[game]": {
      "MapName": "Custom",
      "VsBots": "2:1",
      "Campaign": "No",
      "ForceRespawn": "No",
      "GameMode": "Deathmatch",
      "NumBots": -1,
      "GoalScore": 3,
      "TimeLimit": 2,
      "NumBotsIA": 6,
      "OfficialMutators": [
        "Low Gravity"
      ],
      "SteamVAC": 0,
      "CustomMapName": "DM-1on1-Pure-Pro2",
      "CustomGameMode": "UTGame.UTDeathmatch",
      "CustomMutators": "EasyDodge",
      "CustomMutClasses": "\u001cEasyDodgeMutator\u001c"
    },
    "connect": "192.168.56.56:7777",
    "ping": 23
  }
]
```

The UnrealEngine3 protocol will serialize all provided data. Most of that data is also provided with a master/online query.
UE3 response packet

> [!NOTE]
> Some changes are removed due to a rebase of this PR.

### Changes:
--------
#### Code
- ~~Protocols are now imported and re-exported in `protocols/index.js` for better maintainability~~
- ~~Subsystem protocols are searched by appending it to the given `type` with some common delimiters (`$`, `_`, `:` or nothing) which should be compliant to [naming rule](../../CONTRIBUTING.md#naming)~~
- Some changes regarding the call of `core:run` to provide overloading
- Some changes regarding providing/overloading options

#### Query/Core
- User-Options are prioritized over any default/game/protocol/~~subsystem~~ options
- New method `core:updateOptionsDefaults(..)`, passed with final options to core/protocol to handle given options before the protocol runs
 - Added `core:getOptionsDefaults(outOptions)` to provide default options per protocol
 - Added `core:getOptionsOverrides(outOptions)` to provide overriding options per protocol
- New method `core:prepareRun()` to run before `core:run(state)`
- New method `core:finishRun()` to run after `core:run(state)`
- New method `core:createState()`, for custom handling of creating initial state
- New method `core:finishState()`, for custom handling of finishing the state
- New method `core:populateState()`, for populate the state with queried data (moved from core:run()
- Method `setupOptions(..)` to handle building up options list (should not be overloaded)

#### New Core protocol for LAN-protocols
- Defaults `address` to `255.255.255.255`
- Defaults `givenPortOnly` to `true`
- Enables broadcast mode based on resolved/given address

#### UnrealEngine3/LAN protocol:
Provides basic functionality of parsing UE3 LAN packets send over UDP broadcast
- LAN query data does not contain player data
- UE3 UDP packet definition is compatible to most UE3 based games, some newer engines support Steam which adds an additional 4-byte value (see packetVersion)
- Games can be added by providing `packetVersion` and `gameUniqueId` (+ `port`, see config for _[IpDrv.OnlineGameInterfaceImpl]_ and property _LanAnnouncePort_)
- Some games are using the same LAN port and unique game id (such as UDK SDK or Toxikk), these games need to be differentiated on consuming client)

#### UT3 LAN protocol:
- Added subsystem "LAN", which overrides the query port to be 14001, and sets specific query options for a proper LAN (broadcast) query
- Sets `packetVersion` to `5`, and `gameUniqueId` to `0x4D5707DB`

#### Reader:
- Added method to read 8-byte double values (`double()`)
- Added `peek` parameter for `reader:remaining(..)` to check if the reader can read the additional data

### DIFF
--- a/lib/GlobalUdpSocket.js
+++ b/lib/GlobalUdpSocket.js
@@ -40,8 +40,13 @@ export default class GlobalUdpSocket {
     return this.socket
   }
 
-  async send (buffer, address, port, debug) {
+  async send (buffer, address, port, debug, enableBroadcast = undefined) {
     const socket = await this._getSocket()
+
+    // if broadcast is enabled, pass it to the socket
+    if (enableBroadcast) {
+      socket.setBroadcast(true)
+    }
 
     if (debug) {
       this.logger._print(log => {
--- a/lib/QueryRunner.js
+++ b/lib/QueryRunner.js
@@ -19,6 +19,7 @@ export default class QueryRunner {
       port: runnerOpts.listenUdpPort
     })
     this.portCache = {}
+    this.userOptions = {}
   }
 
   async run (userOptions) {
@@ -28,6 +29,9 @@ export default class QueryRunner {
         userOptions[key] = parseInt(value)
       }
     }
+
+    // cache user options
+    this.userOptions = userOptions
 
     const {
       port_query: gameQueryPort,
@@ -114,8 +118,9 @@ export default class QueryRunner {
 
   async _attempt (options) {
     const core = getProtocol(options.protocol)
-    core.options = options
+    core.options = core.setupOptions(options, this.userOptions)
     core.udpSocket = this.udpSocket
+    core.updateOptionsDefaults()
     return await core.runOnceSafe()
   }
 }
--- a/lib/games.js
+++ b/lib/games.js
@@ -3152,6 +3152,17 @@ export const games = {
       old_id: 'ut3'
     }
   },
+  unrealtournament3lan: {
+    name: 'Unreal Tournament 3',
+    release_year: 2007,
+    options: {
+      port: 14001,
+      port_query_offset: null,
+      protocol: 'ut3lan',
+      packetVersion: 5,
+      lanGameUniqueId: 0x4D5707DB
+    }
+  },
   urbanterror: {
     name: 'Urban Terror',
     release_year: 2000,
--- a/lib/reader.js
+++ b/lib/reader.js
@@ -140,13 +140,27 @@ export default class Reader {
     return r
   }
 
+  double () {
+    let r = 0
+    if (this.remaining() >= 8) {
+      if (this.defaultByteOrder === 'be') r = this.buffer.readDoubleBE(this.i)
+      else r = this.buffer.readDoubleLE(this.i)
+    }
+
+    this.i += 8
+    return r
+  }
+
   varint () {
     const out = Varint.decode(this.buffer, this.i)
     this.i += Varint.decode.bytes
     return out
   }
 
-  /** @returns Buffer */
+  /**
+   * Return the upcoming uffer of given size
+   * @param bytes the amount of bytes to read
+   * @returns Buffer of given size */
   part (bytes) {
     let r
     if (this.remaining() >= bytes) {
@@ -158,14 +172,29 @@ export default class Reader {
     return r
   }
 
-  remaining () {
-    return this.buffer.length - this.i
+  /** Returns the remaining bytes in the current buffer
+   * <b>
+   * Note: remaining can be negative when the current position was exceed due to reading a given sized property
+   * </b>
+   * @param peek? additional bytes to read from the current position and returns the remaining count of bytes (defaults to 0)
+   * @return the remaining byte count
+   */
+  remaining (peek = 0) {
+    return this.buffer.length - this.i - (+peek)
   }
 
+  /**
+   * Returns the remaining byte array from the current position
+   * @returns the remaining bytes
+   */
   rest () {
     return this.buffer.slice(this.i)
   }
 
+  /**
+   * Checks if the buffer has reached its end.
+   * @returns true if the buffer is at the end, otherwise the current position is before the end.
+   */
   done () {
     return this.i >= this.buffer.length
   }
--- a/protocols/core.js
+++ b/protocols/core.js
@@ -18,6 +18,7 @@ export default class Core extends EventEmitter {
     this.delimiter = '\0'
     this.srvRecord = null
     this.abortedPromise = null
+    this.enabledBroadcast = null
     this.logger = new Logger()
     this.dnsResolver = new DnsResolver(this.logger)
 
@@ -253,7 +254,7 @@ export default class Core extends EventEmitter {
     if (typeof buffer === 'string') buffer = Buffer.from(buffer, 'binary')
 
     const socket = this.udpSocket
-    await socket.send(buffer, address, port, this.options.debug)
+    await socket.send(buffer, address, port, this.options.debug, this.enabledBroadcast)
 
     if (!onPacket && !onTimeout) {
       return null
@@ -261,18 +262,27 @@ export default class Core extends EventEmitter {
 
     let socketCallback
     let timeout
+    const results = []
+    const isBroadcast = !!this.enabledBroadcast && (address?.includes('255') ?? false)
 
     try {
       const promise = new Promise((resolve, reject) => {
         const start = Date.now()
         socketCallback = (fromAddress, fromPort, buffer) => {
           try {
-            if (fromAddress !== address || fromPort !== port) return
+            // in case of a configured broadcast address, the received response might come from the same "address"
+            // e.g. responses for UE3 lan queries might be sent as broadcast therefore the address can be the same
+            if ((fromAddress !== address && !isBroadcast) || fromPort !== port) return
+
             this.registerRtt(Date.now() - start)
             const result = onPacket(buffer)
             if (result !== undefined) {
-              this.logger.debug('UDP send finished by callback')
-              resolve(result)
+              // broadcasts may expect multiple respones, store packet and keep waiting for additional responses
+              results.push(result)
+              if (!isBroadcast) {
+                this.logger.debug('UDP send finished by callback')
+                resolve(result)
+              }
             }
           } catch (e) {
             reject(e)
@@ -282,6 +292,12 @@ export default class Core extends EventEmitter {
       })
       timeout = Promises.createTimeout(socketTimeout, 'UDP')
       const wrappedTimeout = Promise.resolve(timeout).catch((e) => {
+        // in case of broadcast query and received responses, don't return as timeout-error.
+        // consider the timeout out with at least one received response as valid
+        if (isBroadcast && !!results.length) {
+          return results
+        }
+
         this.logger.debug('UDP timeout detected')
         if (onTimeout) {
           const result = onTimeout()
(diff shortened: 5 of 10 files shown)


mmorrisontx:
-1 for putting non-standard query output into the main object. all of that should be in raw, or else it becomes part of the promised standard which is impossible to maintain cross-game.
-1 for subsystem. It may as well just be a different game ID if it's a different query protocol. That's probably up to the current maintainers though.
-1 for doing discovery within a game protocol. Gamedig has always been about querying servers that you know about, and discovery was left out of scope. If discovery is needed, I'd argue that it should be an independent method call, totally separate from the existing query api.

RattleSN4K3:
Thanks for your feedback.

> -1 for putting non-standard query output into the main object. all of that should be in raw, or else it becomes part of the promised standard which is impossible to maintain cross-game.

That's applicable in a sec. Good to know.

> -1 for subsystem. It may as well just be a different game ID if it's a different query protocol. That's probably up to the current maintainers though.

If the game-entry (in games.js) corresponds to a client, isn't it the same game ID then, or am I missing something? 

By looking at Minecraft how it is set up, each entry would require a different client to connect to a given game ID (same with GTA 5 what I can see). Whereas with UT3, you are still _bound_ to the game, same client, same connection method even when having the subsystem.

Both response data (for `unrealtournament3` and such `unrealtournament3lan` game ID) would have the same output (except `players`, or `bots`). What's the ideal approach I can follow?

> -1 for doing discovery within a game protocol. Gamedig has always been about querying servers that you know about, and discovery was left out of scope. If discovery is needed, I'd argue that it should be an independent method call, totally separate from the existing query api.

In various tests the LAN servers for UT3 didn't respond to unicast query. That was/is the goal of this feature.

Currently, my testing server running on localhost responds to a unicast using the same subsystem with the option `address: '192.168.56.56'` But since the response wouldn't be parsed properly with the existing protocol(s), the custom handling would be required. Thus the "subsystem" protocol (although it can be any protocol).

Adding support for discovery was just a small change (7b16301). The only critical change in that would be the non-standard response by receiving 1:n server responses.

---

I likely split the settings-resolver part from this pull request since that can be an additional one. How I can continue changing some bits of the PR?

Hornochs:
Having two game entries for the same game, like `unrealtournament3` and `unrealtournament3lan` has pro and cons.

The pro of having two GameIDs is that it’s more visible for the user that there are two different approaches for the game. Which means when user setup gamedig and seeing the gamelist they don‘t have to look deep into the documentation or notes.

Con is that, at least for me, it would be more „clean“ to handle. Having one gameID and a sub parameter makes more sense. On different games are multiple parameters necessary as well, like nadeo with Userlogin.

CosminPerRam:
Same opinion as @mmorrisontx, but to extend some stuff:
> If the game-entry (in games.js) corresponds to a client, isn't it the same game ID then, or am I missing something?

Hope I got your question right, #608 would benefit from exactly what you describe but introducing something like `subsystem` would be a breaking change (can definitely be considered for the next major version, also considering the plans for rcon support, which would generate more GIDs).

> -1 for doing discovery within a game protocol.

Strongly agreed, #484 is similar to this, and it would be something separated from the query api.

> [...] The only critical change in that would be the non-standard response by receiving 1:n server responses.

For now at least, this could be written in the protocol implementation (or let that be a base protocol).
If multiple games needing this appear, we could then move it over.

RattleSN4K3:
I've rebased the pull request's [source branch](https://github.com/RattleSN4K3/node-gamedig/tree/feat/ue3-lan) as I reduced the complexity, based on the given feedback
- fully removed subsystem
- added unique/separate game ID
- removed unrelated code changes
- backup is here [backup/ue3-lan_1](https://github.com/RattleSN4K3/node-gamedig/tree/backup/ue3-lan_1)

I reply to some of your messages. The current codebase for this PR has changed a bit so you guys may need to review the changes. It is still not fully tested with additional games LAN _query_.

tl;dr The current PR will result into such queries:
https://gist.github.com/RattleSN4K3/55958bfc0639c4ce2347f5298cd70c6a

---
 
> mmorrisontx: Gamedig has always been about querying servers that you know about

(1) I thought about that a bit. There is some smaller issue with that regarding providing the data. Even by knowing the server's IP (on a LAN environment), the single query can lead to (receiving multiple or) just a single random respone since you query the IP using the "LanAnnouncePort" which is shared across any instance on the same system. (from the same game).

A LAN setup where a single machine would host several instances of the same game (if supported) could be the norm.

> CosminPerRam: subsystem would be a breaking change

It was opt-in. Out of interest, could you elaborate what could be considered breaking? All existing protocol wouldn't be affected. 

That subsystem logic could be seen as a set of properties, specified on a game level, to be ready to use. For convience it was auto-resolving the subsystem (when set) to a given protocol, which could be seen as breaking.

> CosminPerRam: Strongly agreed, #484 is similar to this, and it would be something separated from the query api.

I see what you guys mean by splitting any _Discovery API_ from _Query_, but having a setup like mentioned in (1), the query would be incomplete since you cannot specifiy a single system on such LAN setup. That would require various filter options (like only responding to given ports). All game instance (of UT3, and various UE3 games) use the same "query" port.

Since the single query call and response for the LAN protocol is the same with having gamedig being set as broadcast, I would guess query would fit for getting data from LAN servers.
Without [setBroadcast](https://nodejs.org/api/dgram.html#socketsetbroadcastflag) the socket will throw an error when using a _broadcast address_.  

When using the unicast address for a network address/machine which may have/run multiple servers, you won't receive additional ones, only one random (although it seems to be the first one bound to the port).

So the goal of this PR i basically fixing the query of single IP:Port on a local network. The discovery of instances on another network address could be seen as bonus 😁
---

I had breaking changes in mind that's the reason of `enabledBroadcast` required to be _truthy_. Only the Core LAN protocol(s) would set this (based on **detecting a broadcast IP**). So it would never change how the _udpSocket_ acts for any existing protocol (or setup, since it is a new field to be set).

But at the end I might not see the broad system to actualy judge how breaking a change could be. 🙃

Hornochs:
> (1) I thought about that a bit. There is some smaller issue with that regarding providing the data. Even by knowing the server's IP (on a LAN environment), the single query can lead to (receiving multiple or) just a single random respone since you query the IP using the "LanAnnouncePort" which is shared across any instance on the same system. (from the same game).

And that is one of the reason that there is the need of a check if the received Package is from the server you query. Querying ingame uses the same system, the gameclient sends the broadcast to the beacon Port. Imagine the following scenario:

- GameDig gets started to query a server
- Just in that time a player sends an broadcast request

Of course the server should sends two times, but it could be an timing issue and GameDig will receive a Answer Package from a different gameserver, or even of a different game if the unique LANID is different 😐

RattleSN4K3:
> And that is one of the reason that there is the need of a check if the received Package is from the server you query. 

The initiated query/lookup by a game or GameDig will only receive-and-process the response(s) meant for each specific client instance (for UE3). That's what the nonce is for which is not static but generated for each client (UE3 engine wide and in this PR).

https://github.com/gamedig/node-gamedig/blob/2ce6690bd917187023c236344726ea5826e7682f/protocols/unrealengine3lan.js#L32-L33

https://github.com/gamedig/node-gamedig/blob/2ce6690bd917187023c236344726ea5826e7682f/protocols/unrealengine3lan.js#L283-L287

Receiving responses from different games than expecting (due to the games having wrongly used game ids) isn't a problem of Gamedig, the broadcast or the current UDP broadcast implementation of this PR.

CosminPerRam:
> Out of interest, could you elaborate what could be considered breaking? All existing protocol wouldn't be affected.

It isn't a breaking change in the sense that the existing protocols wouldn't be affected, but rather than how all the games are currently being handled by some users, for example [Homepage](https://gethomepage.dev/latest/widgets/services/gamedig/) selects their game solely by the GID, introducing this in v5 would leave some entries in being unreachable by them (although this could be solved by also adding GIDs for these?).

podrivo:
@RattleSN4K3 Any updates on this?

Issue #514, opened 2024-01-25
Battlefield Bad Company 2 is odd and confusing


dgibbs64:
**Describe the bug**
The new name change for Battlefield Bad Company 2 from `bfbc2` to `bbc2` is odd and confusing.

I recommend either keeping the old name of bfbc2 which is the known shorthand name of the game (All Battlefield games shorthand are known as bf* for example bf2, bf3, bf2142 etc) or renaming it to something battlefieldbc2 might be more appropriate. I can see that new rules have been created for naming which is great. However in this case it doesn't seem to work and bending the rules might be a good idea in cases like this.

cetteup:
I totally agree that names should be somewhat consistent within a game/franchise

Arma is another weird one as it stands. Just Arma: Armed Assault is "aaa". Arma 2 - Operation Arrowhead is "a2oa". Arma Reforger, however, is "armareforger". Meaning by looking for "arma" in the IDs, you'd find some Arma games - but not all. The sorting also does not help, since different games of a franchise will be sorted to totally different positions due to how different their names are.

I know incorporating franchises into the naming convention is difficult - especially since what is a single game now may become a franchise down the road. But maybe there's a compromise somewhere.

podrivo:
You can still use old ids by adding the `checkOldIDs` when querying servers. More [here](https://github.com/gamedig/node-gamedig/blob/master/MIGRATE_IDS.md).

dgibbs64:
The old id's will eventually go away so that is not really a solution. My suggestion (while v5 is still in test) is to review the game IDs to see if they make sense to humans and appear consistent. If its hard to guess what the game is just by looking at the gameid then there is a problem. Using the new rules as guidelines rather hard rules you must abide by. Otherwise you will end up with some gameids that make no sense to people. @cetteup gave some good examples of this issue.

podrivo:
There was a PR (https://github.com/gamedig/node-gamedig/pull/487) to add game aliases for a more user friendly approach, but the discussion hasn't evolved, as we had to focus on other stuff at the moment.

CosminPerRam:
[...]

PR #643, opened 2024-09-19
feat: add Renegade X support

RattleSN4K3:
Adds support for Renegade X, using master server
Related issues: #629

> [!NOTE]  
> Setup and Response updated to reflect made changes during developing this PR.

### Query single server
Setup:
```js
GameDig.query({
  type: 'renegade10',
  address: '1.0.0.1',
  port: 7778
```

Example response
```json
{
  "name": "[PREFIX] Server Name",
  "map": "CNC-Field",
  "password": false,
  "raw": {
    "Name": "Server Name",
    "NamePrefix": "[PREFIX]",
    "Current Map": "CNC-Field",
    "Bots": 0,
    "Players": 12,
    "Game Version": "Open Beta 5.84.767",
    "Variables": {
      "Mine Limit": 24,
      "bSteamRequired": false,
      "bPrivateMessageTeamOnly": false,
      "bPassworded": false,
      "bAllowPrivateMessaging": true,
      "bRanked": true,
      "Game Type": 1,
      "Player Limit": 64,
      "Vehicle Limit": 20,
      "bAutoBalanceTeams": false,
      "Team Mode": 6,
      "bSpawnCrates": true,
      "CrateRespawnAfterPickup": 35,
      "Time Limit": 50
    },
    "Port": 7778,
    "IP": "1.0.0.1"
  },
  "version": "Open Beta 5.84.767",
  "maxplayers": 64,
  "numplayers": 12,
  "players": [],
  "bots": [],
  "queryPort": 7777,
  "connect": "1.0.0.1:7778",
  "ping": 0
}
```

### Query and receive master list

Setup:
```js
GameDig.query({
  type: 'renegade10master'
```

Example response
```json
{
  "name": "",
  "map": "",
  "password": false,
  "raw": {
    "servers": [
      {
        "name": "Server Name",
        "map": "CNC-Walls",
        "password": false,
        "raw": {
          "Name": "Server Name",
          "Current Map": "CNC-Walls",
          "Bots": 0,
          "Players": 0,
          "Game Version": "Open Beta 5.71.534",
          "Variables": {
            "Mine Limit": 30,
            "bSteamRequired": false,
            "bPrivateMessageTeamOnly": false,
            "bPassworded": false,
            "bAllowPrivateMessaging": true,
            "bRanked": true,
            "Game Type": 1,
            "Player Limit": 24,
            "Vehicle Limit": 11,
            "bAutoBalanceTeams": false,
            "Team Mode": 6,
            "bSpawnCrates": true,
            "CrateRespawnAfterPickup": 30,
            "Time Limit": 0
          },
          "Port": 7777,
          "IP": "1.0.0.1"
        },
        "version": "Open Beta 5.71.534",
        "maxplayers": 24,
        "numplayers": 0,
        "players": [],
        "bots": [],
        "queryPort": 0
      },
      {
        "name": "[PREFIX] Test Server",
        "map": "CNC-Uphill",
        "password": true,
        "raw": {
          "Name": "Test Server",
          "NamePrefix": "[PREFIX]",
          "Current Map": "CNC-Uphill",
          "Bots": 0,
          "Players": 0,
          "Game Version": "Open Beta 5.79.698",
          "Variables": {
            "Mine Limit": 24,
            "bSteamRequired": false,
            "bPrivateMessageTeamOnly": false,
            "bPassworded": true,
            "bAllowPrivateMessaging": true,
            "bRanked": true,
            "Game Type": 1,
            "Player Limit": 32,
            "Vehicle Limit": 10,
            "bAutoBalanceTeams": false,
            "Team Mode": 1,
            "bSpawnCrates": false,
            "CrateRespawnAfterPickup": 30,
            "Time Limit": 60
          },
          "Port": 7777,
          "IP": "1.0.0.2"
        },
        "version": "Open Beta 5.79.698",
        "maxplayers": 32,
        "numplayers": 0,
        "players": [],
        "bots": [],
        "queryPort": 0
      }
    ]
  },
  "version": "",
  "maxplayers": 0,
  "numplayers": 0,
  "players": [],
  "bots": [],
  "connect": "undefined:undefined",
  "ping": 0
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Feat: Satisfactory - Added support (By @Smidy13 #645)
 * Feat: Update Soldat protocol (#642)
 * Feat: TOXIKK (2016) - Added support (#641)
+* Feat: Renegade X (2014) - Added support (#643)
 
 ## 5.1.3
 * Fix: `Deus Ex` using the wrong protocol (#621)
--- a/protocols/index.js
+++ b/protocols/index.js
@@ -37,6 +37,7 @@ import palworld from './palworld.js'
 import quake1 from './quake1.js'
 import quake2 from './quake2.js'
 import quake3 from './quake3.js'
+import renegadex from './renegadex.js'
 import rfactor from './rfactor.js'
 import samp from './samp.js'
 import satisfactory from './satisfactory.js'
@@ -69,7 +70,7 @@ import vintagestory from './vintagestory.js'
 export {
   armagetron, ase, asa, assettocorsa, battlefield, buildandshoot, cs2d, discord, doom3, eco, epic, factorio, farmingsimulator, ffow,
   fivem, gamespy1, gamespy2, gamespy3, geneshift, goldsrc, gtasao, hexen2, jc2mp, kspdmp, mafia2mp, mafia2online, minecraft,
-  minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, rfactor, ragemp, samp,
+  minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, renegadex, rfactor, ragemp, samp,
   satisfactory, soldat, savage2, starmade, starsiege, teamspeak2, teamspeak3, terraria, toxikk, tribes1, tribes1master, unreal2, ut3, valve,
   vcmp, ventrilo, warsow, eldewrito, beammpmaster, beammp, dayz, theisleevrima, xonotic, altvmp, vintagestorymaster, vintagestory
 }
--- a/protocols/index.js
+++ b/protocols/index.js
@@ -38,6 +38,7 @@ import quake1 from './quake1.js'
 import quake2 from './quake2.js'
 import quake3 from './quake3.js'
 import renegadex from './renegadex.js'
+import renegadexmaster from './renegadexmaster.js'
 import rfactor from './rfactor.js'
 import samp from './samp.js'
 import satisfactory from './satisfactory.js'
@@ -70,7 +71,7 @@ import vintagestory from './vintagestory.js'
 export {
   armagetron, ase, asa, assettocorsa, battlefield, buildandshoot, cs2d, discord, doom3, eco, epic, factorio, farmingsimulator, ffow,
   fivem, gamespy1, gamespy2, gamespy3, geneshift, goldsrc, gtasao, hexen2, jc2mp, kspdmp, mafia2mp, mafia2online, minecraft,
-  minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, renegadex, rfactor, ragemp, samp,
+  minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, renegadex, renegadexmaster, rfactor, ragemp, samp,
   satisfactory, soldat, savage2, starmade, starsiege, teamspeak2, teamspeak3, terraria, toxikk, tribes1, tribes1master, unreal2, ut3, valve,
   vcmp, ventrilo, warsow, eldewrito, beammpmaster, beammp, dayz, theisleevrima, xonotic, altvmp, vintagestorymaster, vintagestory
 }
--- a/protocols/renegadex.js
+++ b/protocols/renegadex.js
@@ -1,4 +1,158 @@
 import Core from './core.js'
+// import Ajv from 'ajv'
+// const ajv = new Ajv()
+
+export const MasterServerServerInfoSchema = {
+  type: 'object',
+  required: [
+    'IP',
+    'Port',
+    'Name',
+    'Current Map',
+    'Bots',
+    'Players',
+    'Game Version',
+    'Variables'
+  ],
+  properties: {
+    IP: {
+      type: 'string',
+      format: 'ipv4',
+      description: 'IP of the server'
+    },
+    Port: {
+      type: 'integer',
+      minimum: 0,
+      maximum: 65535,
+      description: 'The port of the server instance to connect to for joining'
+    },
+    Name: {
+      type: 'string',
+      description: 'Name of the server, i.e.: Bob\'s Server.'
+    },
+    NamePrefix: {
+      type: 'string',
+      description: 'A prefix of the server'
+    },
+    'Current Map': {
+      type: 'string',
+      description: 'The current map\'s name the server is running is running'
+    },
+    Players: {
+      type: 'integer',
+      description: 'The number of players connected to the server',
+      minimum: 0
+    },
+    Bots: {
+      type: 'integer',
+      minimum: 0,
+      description: 'The number of bots'
+    },
+    'Game Version': {
+      type: 'string',
+      pattern: '^Open Beta (.*?)?$',
+      description: 'Version of the build of the server'
+    },
+    Variables: {
+      type: 'object',
+      properties: {
+        'Player Limit': {
+          type: 'integer',
+          minimum: 0,
+          description: 'Maximum number of players allowed by this server'
+        },
+        'Time Limit': {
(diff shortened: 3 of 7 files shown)


RattleSN4K3:
ID test fails but I don't think the ID should be forced as `renegade10`.

CosminPerRam:
Unfortunately it needs to be that due to the current GID formatting rules.
We have gotten a bunch of negative feedback regarding it and will most likely be changed in the next major version (see #514 for more info), but for now we've got to hold onto the old system.

xCausxn:
Even though `0` is falsy, I think we need to make sure `password` is either `true` or `false`

RattleSN4K3:
> Even though `0` is falsy, I think we need to make sure `password` is either `true` or `false`

Good catch. I actually wanted to enforce `"0"`/`"1"` and ended up having 0/1. The downside of working on severel protocols in parallel and and not checking the Results output.

RattleSN4K3:
I changed processing `password`. Due to some commit message, I wanted to rebase the PR, but since it was a draft, I see no problem with it (I generel would not prefer doing so).

PR is ready for review.

protocols/renegadex.js
```diff
@@ -0,0 +1,221 @@
...
+    state.map = serverInfo['Current Map']
+    state.password = variables.bPassworded
+
+    state.numplayers = numplayers
+    state.maxplayers = variables['Player Limit'] || 0
+
+    // due to master server not providing bot/player list, and standard result has no bot count, add list with dummy values
+    state.players = Array.from(new Array(numplayers).keys(), (i) => ({ name: `Player #${i + 1}`, raw: {} }))
+    state.bots = Array.from(new Array(numbots).keys(), (i) => ({ name: `Bot #${i + 1}`, raw: {} }))
+
```
CosminPerRam, line 217: In v4 we had players::setNum, which just filled the players field with empty values if the server wouldn't provide them, but it has been removed in https://github.com/gamedig/node-gamedig/pull/389 as you couldn't really determine whether the server provided the values but are as they are presented or filled by that method, please remove these.

CosminPerRam (changes requested):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 5.X.Y
 * Feat: Replaced `punycode` package usage with `url.domainToASCII` (#630).
 * Feat: World of Padman (2007) - Added support (#636)
+* Feat: Renegade X (2014) - Added support (#643)
```
CosminPerRam, line 6: You could also mention yourself here if you'd like (:

protocols/renegadex.js
RattleSN4K3: The information of how many bots are set/playing whould be left out in the standard result, as #389 was only adding the player count. But yeah okay, could be something to improve/extend.

RattleSN4K3: Remove the lines. But also passed the raw server info as output now,

CHANGELOG.md
RattleSN4K3: It's fine without

protocols/renegadex.js
```diff
@@ -0,0 +1,216 @@
...
+  constructor () {
+    super()
+    this.usedTcp = true
+  }
+
+  async run (state) {
+    const servers = await this.request({
+      url: 'https://serverlist-rx.totemarts.services/servers.jsp',
+      responseType: 'json'
+    })
```
CosminPerRam, line 170: Please split this protocol to a master one (`renegadexmaster`) and this one (`renegadex`) to let users to query just the master server to obtain the server list if they'd like to.

Other similar cases of this: 
* [beammpmaster](https://github.com/gamedig/node-gamedig/blob/c77ca299b017619506bdb0680e5809e36364520f/protocols/beammpmaster.js) - [beammp](https://github.com/gamedig/node-gamedig/blob/c77ca299b017619506bdb0680e5809e36364520f/protocols/beammp.js)
* [vintagestorymaster](https://github.com/gamedig/node-gamedig/blob/c77ca299b017619506bdb0680e5809e36364520f/protocols/vintagestorymaster.js) - [vintagestory](https://github.com/gamedig/node-gamedig/blob/c77ca299b017619506bdb0680e5809e36364520f/protocols/vintagestory.js)
* [tribes1master](https://github.com/gamedig/node-gamedig/blob/c77ca299b017619506bdb0680e5809e36364520f/protocols/tribes1master.js) - note that [tribes1](https://github.com/gamedig/node-gamedig/blob/c77ca299b017619506bdb0680e5809e36364520f/protocols/tribes1.js) does not use the master server for the query, but we still have the master server implementation.

CosminPerRam (changes requested):

protocols/renegadex.js
RattleSN4K3: That master-separation doesn't seem to be fully implemented (respectively exposed for the API). The master query logic is also a bit inconsistent throughout the protocols (see `altvml` and `ragemp` [or `minetest`]).

Also by having a master protocol (like `vintagestorymaster`) there could be the assumption of receiving a list of processed server data and not just the raw master server data. And vice versa

Not sure how I feel about that. Should I still follow the mentioned split-master-logic?

CosminPerRam: I'd say yes, although we don't really mention anything about the support of querying master servers (and that it is flaky right now) we can provide it as just it is right now and straighten it up for the next major release (we should do a documented issue about this).

RattleSN4K3: Added that master protocol. I can't tell if I like the current logic of adding master-protocols (the earlier subsystem I propsoed in #638 could be a alternative), but here we go.

The master server can be used wih `protocol-renegadexmaster` (or to be consistent to current game ID naming, `protocol-renegade10master`. The server list will be processed to not provide different type of data.

protocols/index.js
```diff
@@ -37,6 +37,7 @@ import palworld from './palworld.js'
 import quake1 from './quake1.js'
 import quake2 from './quake2.js'
 import quake3 from './quake3.js'
+import renegadex, { renegadexmaster } from './renegadex.js'
```
CosminPerRam, line 40: Please split the master protocol query in another file.

protocols/index.js
```diff
@@ -67,7 +68,7 @@ import soldat from './soldat.js'
 export {
   armagetron, ase, asa, assettocorsa, battlefield, buildandshoot, cs2d, discord, doom3, eco, epic, factorio, farmingsimulator, ffow,
   fivem, gamespy1, gamespy2, gamespy3, geneshift, goldsrc, gtasao, hexen2, jc2mp, kspdmp, mafia2mp, mafia2online, minecraft,
-  minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, rfactor, ragemp, samp,
+  minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, renegadex, renegadexmaster, renegadexmaster as renegade10master, rfactor, ragemp, samp,
```
CosminPerRam, line 71: Export only one entry, which should be named as `${game_id}master`, so `renegadexmaster`.
(side note: this is also not specified anywhere nor standalized unfortunately)

CosminPerRam (changes requested):

protocols/index.js
RattleSN4K3: That's the issue as the _game id_ is forced to be `...10`. `renegadex...` would be wrong to due to  the strict _normalisation_. Exporting the "normalized" name was just a shortcut solution.

protocols/index.js
RattleSN4K3: I would have thought protocols (as file) provide the logic for the single protocol. Thus as both classes are using the master-server, I kept it in the same file.

With possible changes (for instance with #608) the file structured could be scattered.

RattleSN4K3:
Updated the PR to reflect the proposed changes.

My points regarding the last changes still stand tho:
- Alternative protocols in a single file could be messy... in case of protocols being dependet on existing ones (see `master` and server-query based on `master` protocol)
  > With possible changes (for instance with #608) the file structured could be scattered.
- Inconsistency in protocol names and game id

CosminPerRam (approved):

Issue #442, opened 2023-12-29
Support Satisfactory game


Maxence1502:
Hello,

Is it possible to support this game : https://store.steampowered.com/app/526870/Satisfactory/

Thank you :)

CosminPerRam:
Hello, please use the issue templates.
It seems that Satisfactory doesn't support valve server queries, but I've found a mod thats made for this: [FicsIt Remote Monitoring](https://ficsit.app/mod/FicsitRemoteMonitoring).
I've seen on the [steamdb page of the dedicated server](https://steamdb.info/app/1690800/) that it uses Epic Online Services, it could be queried that way, but finding the necessary keys for the epic protocol is tricky  here.

CosminPerRam:
None that I know of at least unfortunately :/

CosminPerRam:
From their [patchnotes](https://www.reddit.com/r/SatisfactoryGame/comments/1fdjz13/patch_notes_satisfactory_10_build_365306/):
> We have added support for an HTTP API, so it is now possible to create applications to communicate with the dedicated server for those who are interested in developing tools for it :) 

And upon starting a dedicated server (hope it means what I think):
> PacketHandlerLog: Loaded PacketHandler component: FactoryDedicatedServer.DSQueryHandlerComponentFactory (DSQueryHandlerComponent)

So I guess its possible, I'll look into it later on.

CosminPerRam:
From their Discord server:
![image](https://github.com/user-attachments/assets/941d6bc7-bcfe-483f-a335-dd274ed40f63)

I unfortunately do not own the game.

Destarianon:
[DedicatedServerAPIDocs.md](https://github.com/user-attachments/files/16957580/DedicatedServerAPIDocs.md)
Here is a reference to the API documentation.

Maxence1502:
Let me know if you need help for testing or debugging in game

Smidy13:
Investigated this and will look at contributing. There are two requests needed, one to get a authentication token and then the second to get the server state. Making some notes here and will continue tomorrow.

**First Request to Get Authorization Token**
```
> POST https://localhost:7777/api/v1
> Content-Type: application/json
> data {"function":"PasswordlessLogin","data":{"MinimumPrivilegeLevel":"Client"}}
```

This will return a JSON with an authenticationToken.

**Second Request to Get Server State**
```
POST https://localhost:7777/api/v1
> Content-Type: application/json
> Authorization: Bearer **authenticationToken**
[...]

PR #645, opened 2024-09-24
feat: Added Satisfactory Protocol and Support

Smidy13:
This PR adds support for Satisfactory as requested in https://github.com/gamedig/node-gamedig/issues/442

Please note this is my first PR to this project so please be kind.

Also Satisfactory has only just gone v1.0 therefore the API is limited at the moment. Server Name for example cannot be pulled using the API but a feature request has been submitted to the game developers here which should be voted on for consideration:

https://questions.satisfactorygame.com/post/66ebebad772a987f4a8b9ef8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 5.X.Y
 * Feat: Replaced `punycode` package usage with `url.domainToASCII` (#630).
 * Feat: World of Padman (2007) - Added support (#636)
+* Feat: Satisfactory - Added support (#442) (By @Smidy13)
 
 ## 5.1.3
 * Fix: `Deus Ex` using the wrong protocol (#621)
--- a/GAMES_LIST.md
+++ b/GAMES_LIST.md
@@ -258,6 +258,7 @@
 | rust                 | Rust                                             | [Valve Protocol](#valve)                        |
 | s2ats                | Savage 2: A Tortured Soul                        |                                                 |
 | sdtd                 | 7 Days to Die                                    | [Valve Protocol](#valve)                        |
+| satisfactory         | Satisfactory                                     |                                                 |
 | serioussam           | Serious Sam                                      |                                                 |
 | serioussam2          | Serious Sam 2                                    |                                                 |
 | shatteredhorizon     | Shattered Horizon                                | [Valve Protocol](#valve)                        |
--- a/protocols/index.js
+++ b/protocols/index.js
@@ -39,6 +39,7 @@ import quake2 from './quake2.js'
 import quake3 from './quake3.js'
 import rfactor from './rfactor.js'
 import samp from './samp.js'
+import satisfactory from './satisfactory.js'
 import savage2 from './savage2.js'
 import starmade from './starmade.js'
 import starsiege from './starsiege.js'
@@ -67,7 +68,7 @@ export {
   armagetron, ase, asa, assettocorsa, battlefield, buildandshoot, cs2d, discord, doom3, eco, epic, factorio, farmingsimulator, ffow,
   fivem, gamespy1, gamespy2, gamespy3, geneshift, goldsrc, gtasao, hexen2, jc2mp, kspdmp, mafia2mp, mafia2online, minecraft,
   minecraftbedrock, minecraftvanilla, minetest, mumble, mumbleping, nadeo, openttd, palworld, quake1, quake2, quake3, rfactor, ragemp, samp,
-  savage2, starmade, starsiege, teamspeak2, teamspeak3, terraria, tribes1, tribes1master, unreal2, ut3, valve,
+  satisfactory, savage2, starmade, starsiege, teamspeak2, teamspeak3, terraria, tribes1, tribes1master, unreal2, ut3, valve,
   vcmp, ventrilo, warsow, eldewrito, beammpmaster, beammp, dayz, theisleevrima, xonotic, altvmp, vintagestorymaster,
   vintagestory
 }
--- a/protocols/satisfactory.js
+++ b/protocols/satisfactory.js
@@ -1,0 +1,92 @@
+import Core from './core.js'
+
+/**
+ * Satisfactory servers run by default self-signed certificates. This is used to ignore the validation for the requests.
+ */
+const https = {
+  rejectUnauthorized: false
+}
+
+export default class satisfactory extends Core {
+  constructor () {
+    super()
+
+    /**
+     * To get information about the Satisfactory game server, you need to first obtain a client authenticationToken.
+     * https://satisfactory.wiki.gg/wiki/Dedicated_servers/HTTPS_API
+     */
+
+    this.authenticationToken = null
+
+    // Don't use the tcp ping probing
+    this.usedTcp = true
+
+  }
+
+  async run (state) {
+    await this.getClientAuthenticationToken()
+
+    await this.queryInfo(state)
+    await this.cleanup(state)
+  }
+
+  async getClientAuthenticationToken () {
+    this.logger.debug('Requesting client access token ...')
+
+    const url = `https://${this.options.host}:${this.options.port}/api/v1/`
+
+    const json = {
+      function: 'PasswordlessLogin',
+      data: {
+        MinimumPrivilegeLevel: 'Client'
+      }
+    }
+
+    const headers = {
+      'Content-Type': 'application/json'
+    }
+
+    this.logger.debug(`POST: ${url}`)
+    const response = await this.request({ url, json, https, headers, method: 'POST', responseType: 'json' })
+
+    if (response.data == null) {
+      throw new Error('Unable to retrieve authenticationToken')
+    }
+
+    this.authenticationToken = response.data.authenticationToken
+  }
+
+  async queryInfo (state) {
+    const url = `https://${this.options.host}:${this.options.port}/api/v1/`
+
+    const json = {
+      function: 'QueryServerState'
+    }
+
+    const headers = {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${this.authenticationToken}`
+    }
+
+    this.logger.debug(`POST: ${url}`)
+    const response = await this.request({ url, json, https, headers, method: 'POST', responseType: 'json' })
+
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 5.X.Y
 * Feat: Replaced `punycode` package usage with `url.domainToASCII` (#630).
 * Feat: World of Padman (2007) - Added support (#636)
+* Feat: Satisfactory - Added support (#442) (By @Smidy13)
```
CosminPerRam, line 6: Just a nitpick, please use the format of `(By {name} #{pr})`, as in `(By @Smidy13 #442)`.

protocols/satisfactory.js
```diff
@@ -0,0 +1,92 @@
...
+    this.usedTcp = true
+
+  }
+
+  async run (state) {
+    await this.getClientAuthenticationToken()
+
+    await this.queryInfo(state)
+    await this.cleanup(state)
+  }
```
CosminPerRam, line 31: Please do
```
const authenticationToken = await this.getAuthToken();
await this.queryInfo(state, authenticationToken);
```
to avoid having the `cleanup` function just for a single variable.

protocols/satisfactory.js
```diff
@@ -0,0 +1,92 @@
+import Core from './core.js'
+
+/**
+ * Satisfactory servers run by default self-signed certificates. This is used to ignore the validation for the requests.
+ */
+const https = {
+  rejectUnauthorized: false
+}
```
CosminPerRam, line 8: I'd have to look again through the docs, but if you can provide trusted certificaties in a server we wouldn't want this to be hardcoded, instead, set this value only if it is not provided (for example as in `if (!this.options.port) this.options.port = 27016`).

protocols/satisfactory.js
```diff
@@ -0,0 +1,92 @@
...
+        MinimumPrivilegeLevel: 'Client'
+      }
+    }
+
+    const headers = {
+      'Content-Type': 'application/json'
+    }
+
+    this.logger.debug(`POST: ${url}`)
+    const response = await this.request({ url, json, https, headers, method: 'POST', responseType: 'json' })
```
CosminPerRam, line 50: The difference between this chunk and the one for `queryInfo` is just the json and headers, we can make a factory function for it to avoid code duplication.

CosminPerRam (changes requested):
Looks good, just some small changes, thanks!

protocols/satisfactory.js
Smidy13: I'm a bit stuck on this one - there is an option in the Satisfactory API where you can provide your own certificate for the REST API. This is documented in the Satisfactory API docs here:
https://satisfactory.wiki.gg/wiki/Dedicated_servers/HTTPS_API#Certificate_Validation_and_Encryption

node-gamedig uses the Got HTTP library under the surface for making HTTP requests. This is limited in that there is the rejectUnauthorized flag which will allow self-signed or expired certificates. There is the checkServerIdentity function, but this function is not available for self-signed certificates (tested extensively and it doesn't run).

https://github.com/sindresorhus/got/blob/main/documentation/5-https.md

I think there are couple of different options:

1. We remove the rejectUnauthorized flag and document node-gamedig doesn't support self-signed certificates for Satisfactory and will only connect if the user has defined a valid certificate for the Satisfactory API.

2. We continue to set the rejectUnauthorized flag to false noting and document the use of self-signed certificates are a security risk. 

3. We implement an socket listener or similar to check the certificate as the connection is open (by the time the protocol file is dealing with it the connection is closed and certificate information is not available). If we do this we could perform some checks around certificate validity (i.e. certificate is not expired, name matches expected values). This might require more libraries to achieve and involved messing with the core.js code which I'm less comfortable around.

CHANGELOG.md
Smidy13: Updated to show PR instead of issue reference.

protocols/satisfactory.js
Smidy13: Have updated as requested as well as developing a function for the requests as per other noted comment.

protocols/satisfactory.js
CosminPerRam: Love it!

protocols/satisfactory.js
CosminPerRam: 1: I wouldn't go for this one as I think many users will not setup proper certificates.
3: This one neither, it involves changing quite a lot of structure and I don't quite see the benefits just yet.
2: We could do this considering the fact that many other (HTTP) protocols aren't secured, but we can still go with an `if (!this.options.rejectUnauthorized) this.options.rejectUnauthorized = false` (and add a note about security and this option to disable this behaviour) to make sure the users have control over this.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 5.X.Y
 * Feat: Replaced `punycode` package usage with `url.domainToASCII` (#630).
 * Feat: World of Padman (2007) - Added support (#636)
+* Feat: Satisfactory - Added support (#645) (By @Smidy13)
```
CosminPerRam, line 6: Please do this format for mentioning the PR and author: `(By @author #645)`

CosminPerRam (changes requested):
Also please rebase.

protocols/satisfactory.js
Smidy13: Have gone with Option #3 and documented. I thought about whether to add this to the core.js request but instead chose to set it at the protocol level. Reasoning being is future protocols may wish to use other flags in the https options so better to set at the lower level.

Smidy13:
Hopefully should be OK now. Also thank you for your patience and guidance as I've worked through the options and changes!

README.md
```diff
@@ -62,6 +62,7 @@ Confused on how this works, or you want to see more? Checkout the [examples](/ex
 | **portCache**              | boolean | true      | After you queried a server, the second time you query that exact server (identified by specified ip and port), first add an attempt to query with the last successful port.                                                   |
 | **noBreadthOrder**         | boolean | false     | Enable the behaviour of retrying an attempt X times followed by the next attempt X times, otherwise try attempt A, then B, then A, then B until reaching the X retry count of each.                                           |
 | **checkOldIDs**            | boolean | false     | Also checks the old ids amongst the current ones.                                                                                                                                                                             |
+| **rejectUnauthorized**     | boolean | true      | Specifies whether connections using invalid certificates (i.e. self-signed certificates) are rejected.                                                                                                                        |
```
CosminPerRam, line 65: Games that require additional arguments are not to be listed here, only in the games list file.

lib/QueryRunner.js
```diff
@@ -10,7 +10,8 @@ const defaultOptions = {
   portCache: true,
   noBreadthOrder: false,
   ipFamily: 0,
-  requestPlayers: true
+  requestPlayers: true,
+  rejectUnauthorized: true,
```
CosminPerRam, line 14: We can remove this as it isn't a general option, the default value will be set inside the protocol.

protocols/index.js
```diff
@@ -63,13 +65,11 @@ import xonotic from './xonotic.js'
 import altvmp from './altvmp.js'
 import vintagestorymaster from './vintagestorymaster.js'
 import vintagestory from './vintagestory.js'
-import soldat from './soldat.js'
```
CosminPerRam, line 66: Nitpick: why did this moved? eslint?

protocols/satisfactory.js
```diff
@@ -0,0 +1,75 @@
...
+     */
+
+    state.numplayers = queryResponse.data.serverGameState.numConnectedPlayers
+    state.maxplayers = queryResponse.data.serverGameState.playerLimit
+    state.raw = queryResponse
+
+  }
+
+  async queryInfo (json, headers) {
+
```
CosminPerRam, line 52: Please do 
```js
if (!this.options.rejectUnauthorized) this.options.rejectUnauthorized = false
```

CosminPerRam (changes requested):
Looks good but I've requested some changes to be more in line with the other implementations regarding additional options.

README.md
Smidy13: Removed commentary from README.md but left in place in GAMES_LIST.md.

lib/QueryRunner.js
Smidy13: Removed defaulting in QueryRunner.js and added default parameter in protocol file.

protocols/index.js
Smidy13: Apologies, I had the automated checks state the GAMES_LIST.md needed to be in alphabetical order. The this was just me tidying the alphabetical order in the index.js so that soldat is in alphabetical order for consistency.

protocols/satisfactory.js
Smidy13: Added in to protocol file as specified. Also included a comment referencing the reason for this.

GAMES_LIST.md
```diff
@@ -492,5 +493,9 @@ EOS does not provide players data.
 Palworld support can be unstable, the devs mention the api is currently experimental.  
 To query Palworld servers, the `RESTAPIEnabled` setting must be `True` in the configuration file, and you need to pass the `username` (currently always `admin`) and the `adminpassword` (from the server config) as the `password` parameter.
 
+### <a name="satisfactory"></a>Satisfactory
+Satisfactory servers unless specified use self-signed certificates for the HTTPS API. Because of this you will need to set the `rejectUnauthorized` flag in options to `false` in order to connect.
```
CosminPerRam, line 497: `false` should be `true` here.

CosminPerRam (changes requested):
The CI fails cause we need to have
```
extra: {
   doc_notes: 'satisfactory',
}
```
in the game entry for the markdown file to properly link to its section, for example this is the same for `Grand Theft Auto V - RageMP`:
https://github.com/gamedig/node-gamedig/blob/39d7f06bd9a49008c4f7a224d02e739406677ab2/lib/games.js#L1289

By the way, one can use [act](https://nektosact.com/) to run the github actions locally to avoid waiting for PR reviewer to approve them.

CosminPerRam (approved):
Thank you!

PR #652, opened 2024-10-12
feat: Satisfactory add lightweight query

CosminPerRam:
#645 added support but the current state of the query response is not that great, it does not provide even the server name or version, this PR makes use of the lightweight query API (which is made via socket packets) to get these information.


CosminPerRam:
Tested on a few community-provided servers and my local instance, seems all good on the lightweight part.

PR #661, opened 2025-01-01
feat: add Exfil support

### LINKED ISSUE #659: Please add support for Exfil
Just looking to see if you can add support for Exfil servers. 

Here is a server IP and query port to help test: 
63.143.62.91:7778

Let me know if you need more information.

CosminPerRam:
Closes #659

This is a recently released game, I have confirmed that this works on the issue author's server but I couldn't find anything related to default ports, will merge when so.


CosminPerRam:
I found out they have a Discord server, where they have channels for dedicated servers, where I found this guide, which tells `27015` is the default query port.
[dedicated-windows.md](https://github.com/user-attachments/files/18482911/dedicated-windows.md)

PR #663, opened 2025-01-04
fix: Use cp tag for accurate rust player count

xCausxn:
Rust seems to report numplayers incorrectly over A2S. However the tag system exposes the correct number via the cp tag.

GameQ also implements this logic and we have done as well for max players.


CosminPerRam (approved):
Thanks!

PR #671, opened 2025-01-24
Adding Trackmania 2020 as supported Game

Hornochs:
After investing so much time and testing I forgot to add in my last PR that Trackmania 2020 is also supported

### DIFF
--- a/lib/games.js
+++ b/lib/games.js
@@ -3067,6 +3067,18 @@ export const games = {
       doc_notes: 'nadeo'
     }
   },
+  trackmania2020: {
+    name: 'Trackmania 2020',
+    release_year: 2020,
+    options: {
+      port: 2350,
+      port_query: 5000,
+      protocol: 'nadeo'
+    },
+    extra: {
+      doc_notes: 'nadeo'
+    }
+  },
   trackmaniaforever: {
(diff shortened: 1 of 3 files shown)


lib/games.js
```diff
@@ -3067,6 +3067,18 @@ export const games = {
       doc_notes: 'nadeo'
     }
   },
+  trackmania2020: {
+    name: 'Trackmania 2020',
```
CosminPerRam, line 3071: The game name can be just "trackmania", as the original game title does not include the year in the name.

CosminPerRam (changes requested):
LGTM, please also add a line in the CHANGELOG file

lib/games.js
Hornochs: But then it wouldn‘t be possible to difference between this Version and the OG Trackmania. In the Community the Term 2020 is established

Hornochs:
Files Updated 🫡

podrivo:
@Hornochs Any updates on your side?

@CosminPerRam Does it look ok to merge?

podrivo:
Closes [672](https://github.com/gamedig/node-gamedig/issues/672)

PR #678, opened 2025-02-26
fix(protocols/minetest): rename ip to address

hjri:
Minetest changed property name in list


CosminPerRam (commented):
Thanks for the fix, please also add a line in the CHANGELOG (if you want to, I could do it), something like this:
`Fix: Minetest - server filtering replace missing field ip with address (By @hjri #678)`

PR #686, opened 2025-04-20
fix(protocols/terraria): add missing maxplayers (and raw) to terraria protocol

GoodDays13:
I noticed that maxplayers was 0 for me and it seems that it wasn't being set.


CosminPerRam (commented):
Thanks for the fix!
Would you like to also add a line in the `CHANGELOG.md` file?
Example of line: 
`* Fix: Terraria - add missing supported maxplayers field (By @GoodDays13 #686)`

CosminPerRam (approved):
Thanks!

PR #710, opened 2025-06-07
fix: ignore cached ports when givenPortOnly is set

### LINKED ISSUE #709: bug: port cache adds extra attempts when using `givenPortOnly`
**Describe the bug**
Even when using the `givenPortOnly` option, extra attemps are added when the port cache is enabled.

**Steps To Reproduce**
Using the latest version, run two queries to the same server back to back with the port cache enabled and `givenPortOnly` set to `true`. The first query needs to pass, the second one needs to fail. The easiest way to enforce this is to use an impossible timeout for the second query. Because the port cache is populated for the second query, it will fail with "Failed all 2 attempts".

```js
import { GameDig } from '../lib/index.js'

await GameDig.query({
  type: 'battlefield3',
  host: '79.235.227.110',
  port: 47200,
  givenPortOnly: true,
  maxRetries: 0,
  socketTimeout: 2000,
  attemptTimeout: 2000,
  debug: true,
}).then((state) => {
  console.log(state)
}).catch((error) => {
  console.log(`Server is offline, error: ${error}`)
})

await GameDig.query({
  type: 'battlefield3',
  host: '79.235.227.110',
  port: 47200,
  givenPortOnly: true,
  maxRetries: 0,
  socketTimeout: 10, // impossible timeout, no (remote) server will respond within 10 milliseconds.
  attemptTimeout: 10,
  debug: true,
}).then((state) => {
  console.log(state)
}).catch((error) => {
  console.log(`Server is offline, error: ${error}`)
})
```

**Expected behavior**
**Only** attempt to query the server on the given port if `givenPortOnly` is `true`.

**Additional context**
I am aware that the port cache can simply be disabled, but I find the current behavior confusing. Getting "Failed all 2 attempts" when I specifically asked GameDig to only make an attempt using the port I pass isn't what I personally would expect.

cetteup:
closes #709


xCausxn:
Will need a changelog item though

xCausxn:
Yeah, just thought I would be explicit as I did send a LGTM :)

CosminPerRam (approved):

PR #719, opened 2025-07-21
feat: replace eslint with standardjs

CosminPerRam:
This is experimental for now.

TODO: add CI check


CosminPerRam:
I'm not exactly too fond of it, so I'll close this for now.

PR #725, opened 2025-07-31
Fix: Add User-Agent header to Vintage Story master server request

nunopaiva1:
The Vintage Story master server returns HTTP 403 unless a valid User-Agent is provided.

This change adds a proper User-Agent header to allow successful querying.


protocols/vintagestorymaster.js
```diff
@@ -1,17 +1,23 @@
...
+      responseType: 'json',
+      headers: {
+        'User-Agent': USER_AGENT
+      }
     })
 
     state.raw.servers = response?.data || []
   }
-}
+}
```
CosminPerRam, line 23: Please add a end file new line

CosminPerRam (changes requested):
Please also add a line in the CHANGELOG.md file.

tyronx:
This has been blocked by us on purpose because users of this software are completely overloading our services. It is being spammed to no end. To whoever uses this software, please refrain from using our online service to get your own servers player count - read it out from the `/stats` command.

tyronx:
we don't support rcon (yet). You need to read it out from stdout. the screen command can do that

podrivo:
@nunopaiva1 Were you able to work on an alternative solution as mentioned?

@tyronx Would you mind giving more details on the `/stats` command you mentioned?

Thank you!

podrivo:
@CosminPerRam Any other things we have to consider here or is it good to merge?

PR #731, opened 2025-08-29
feat: update squad support to eos

### LINKED ISSUE #545: bug: Squad now using EOS instead of Valve
**Describe the bug**
It appears that Squad is now using EOS for server management and valve protocol appears to no longer be working. I dont have much experiance in the EOS system yet so unsure how to test it currently.

k3ithos:
Squad has used EOS for a while now so the old Valve protocol implementation no longer works. OAuth credentials extracted by intercepting game traffic

Closes #545

### DIFF
--- a/protocols/squad.js
+++ b/protocols/squad.js
@@ -14,6 +14,7 @@ export default class squad extends Epic {
   async run (state) {
     await super.run(state)
     state.name = state.raw.attributes.SERVERNAME_s
+    state.password = state.raw.attributes.PASSWORD_b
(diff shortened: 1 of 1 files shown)


CosminPerRam:
Oops, the CI failed on generating the games list as we needed to add Squad in the `tools/generate_games_list` blanket EOS inclusion list, I added this in https://github.com/gamedig/node-gamedig/commit/f715b7fe5cf0b13fdea6a6b33d51e056a6963c2b

PR #732, opened 2025-08-30
fix: corrected star trek elite force 2 name

dgibbs64:
I noticed that Star Trek: Elite Force II is named incorrectly, this PR corrects the name.

Star Trek: Voyager - Elite Force 2 > Star Trek: Elite Force II

https://en.wikipedia.org/wiki/Star_Trek:_Elite_Force_II

### DIFF
--- a/lib/games.js
+++ b/lib/games.js
@@ -2812,8 +2812,8 @@ export const games = {
       protocol: 'quake3'
     }
   },
-  stvef2: {
-    name: 'Star Trek: Voyager - Elite Force 2',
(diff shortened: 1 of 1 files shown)


podrivo:
Thanks for the PR, @dgibbs64!

I think we're just missing an update on the `GAMES_LIST.md`.
If you run `node tools/generate_games_list.js` it should do the trick!

podrivo:
@CosminPerRam Should we also consider adding the previous id as an `old_id`?

Issue #702, opened 2025-05-25
bug: Minecraft - {"error":"Failed all 1 attempts"}


dudekm:
```
gamedig --type minecraft minecraft.pl
{"error":"Failed all 1 attempts"}
```
or
```
gamedig --type minecraft minecraft.pl:25565
{"error":"Failed all 2 attempts
```

but here everything is ok:
https://mcstatus.io/status/java/minecraft.pl
https://mcstatus.io/status/bedrock/minecraft.pl

xCausxn:
It looks like that site is just resolving the SRV record potentially.

The server is actually on a different port from the looks of it:

https://gamedig.xcausxn.com/minecraft/minecraft.pl:25716

dudekm:
Fix https://github.com/gamedig/node-gamedig/pull/753

miaopan607:
I met the same problem after updating my uptime-kuma

PR #753, opened 2026-02-06
SRV fix & tests

dudekm:
https://github.com/gamedig/node-gamedig/issues/702


podrivo:
Thanks for the PR, @dudekm!

And I was wondering, does it make more sense to have it separated in two PRs? One for the fix and another one for tests. That way we separate conversations!

CosminPerRam:
Thank you for submitting the changes, but having the 3 line fix change + test suite + node upgrade in a single PR with also no explanation/motive at all is literally not mergeable.

The 3-line fix can come in a scoped small PR.
The test suite is not that easy to merge with no explaination.
The node version upgrade requires a major version release, which we are not ready to do (also without discussing if we should do so at all or what it brings new), also making the fix not available to the previous major without backporting.

Feel free to open a new PR for the SRV fix, thanks.

PR #764, opened 2026-04-04
chore: update tools README

podrivo:
Adding an updated README containing a brief explanation for the existing tools. 
Wasn't too sure about the content, so feel free to edit as needed.


podrivo:
@CosminPerRam Does this makes sense to you? Thoughts?

